PR #3, opened 2015-09-22
Unadvertise should drain before close

anson627:
Today, a peer and its associated connections are closed immediately after unadvertisement request is made, so the request itself fails with a connection reset error.

Now since we have graceful drain, we can decline new requests and wait for on-going requests (including unadvertisement) to finish before closing the connections.

r: @Raynos @ShanniLi @jcorbin

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -354,17 +354,17 @@ function removeServicePeer(serviceName, hostPort) {
     }
 
     if (!anyOtherSubChan) {
-        var connections = peer.connections;
-        var allDrained = CountedReadySignal(connections.length);
+        var allDrained = CountedReadySignal(peer.connections.length + 1);
         allDrained(onAllDrained);
-        for (var j = 0; j < connections.length; j++) {
-            connections[j].drain('closing due to unadvertisement', allDrained.signal);
+        for (var j = 0; j < peer.connections.length; j++) {
+            peer.connections[j].drain('closing due to unadvertisement', allDrained.signal);
         }
+        allDrained.signal();
     }
 
     function onAllDrained() {
-        self.logger.info('Peer drained and closed', self.extendLogInfo({
-            hostPort: peer.hostPort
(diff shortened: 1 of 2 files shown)


service-proxy.js
```diff
@@ -353,6 +354,18 @@ function removeServicePeer(serviceName, hostPort) {
     }
 
     if (!anyOtherSubChan) {
+        var connections = peer.connections;
```
jcorbin: no need for this alias, just peer.connections please

service-proxy.js
```diff
@@ -353,6 +354,18 @@ function removeServicePeer(serviceName, hostPort) {
...
+        var allDrained = CountedReadySignal(connections.length);
```
jcorbin: make this use the N+1 pattern:
```
ready = CountedReadySignal(N+1);
for each thing, doit(ready.signal)
ready.signal(); // this is the +1 call that guarantees that you don't get an erroneous ready signal fire mid for loop above
```

service-proxy.js
```diff
@@ -353,6 +354,18 @@ function removeServicePeer(serviceName, hostPort) {
...
+        allDrained(onAllDrained);
+        for (var j = 0; j < connections.length; j++) {
+            connections[j].drain('closing due to unadvertisement', allDrained.signal);
+        }
+    }
+
+    function onAllDrained() {
+        self.logger.info('Peer drained and closed', self.extendLogInfo({
```
jcorbin: should use peer.extendLogInfo, but that doesn't exist yet... and now the repos are separated, so we have to add it first before we use it here...

Raynos: `"Peer drained and closed due to advertisement"`

service-proxy.js
kriskowal: I know you can’t see this, but I’m stifling a cough.

Raynos:
legit.

Raynos:
Has no method `extendLogInfo`.

Please update tchannel.

PR #5, opened 2015-09-23
Partial affinity

kriskowal:


service-proxy.js
```diff
@@ -320,11 +327,73 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    peer.connectTo();
+};
+
+ServiceDispatchHandler.prototype.refreshServicePeerPartially =
+function refreshServicePeerPartially(serviceName, hostPort) {
+    var self = this;
 
+    // Reset the expiration time for this service peer
     var time = self.channel.timers.now();
     self.exitServices[serviceName] = time;
```
kriskowal: restore in non-enabled case

service-proxy.js
```diff
@@ -320,11 +327,73 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
 function refreshServicePeer(serviceName, hostPort) {
     var self = this;
 
+    if (self.partialAffinityEnabled) {
+        return self.refreshServicePeerPartially(serviceName, hostPort);
+    }
+
+    // The old way: fully connect every egress to all affine peers.
     var peer = self.getServicePeer(serviceName, hostPort);
-    peer.connect(true);
```
kriskowal: separate into other diff

service-proxy.js
```diff
@@ -320,11 +327,73 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+    // Obtain and sort the affine worker and relay lists.
+    var relays = Object.keys(self.egressNodes.exitsFor(serviceName));
+    relays.sort();
+    var serviceChannel = self.getOrCreateServiceChannel(serviceName);
+    var workers = serviceChannel.peers.keys();
+    workers.sort();
+
+    // Find our position within the affine relay set so we can project that
+    // position into the affine worker set.
+    var relayIndex = search(relays, self.channel.hostPort);
```
kriskowal: or binaryIndexOf

kriskowal:
r @Raynos

test/partial-affinity.js
```diff
@@ -0,0 +1,146 @@
...
+    seedConfig: {
+        'hyperbahn.partialAffinity': {
+            minPeersPerRelay: 1,
+            minPeersPerWorker: 1
+        }
+    }
+}, function t(cluster, assert) {
+    runScenario(cluster, assert, onCompletion);
+    function onCompletion(err) {
+        assert.end(err);
```
Raynos: in all cases please assert(numberOfExpectedSockets, N)

service-proxy.js
```diff
@@ -77,6 +80,10 @@ function ServiceDispatchHandler(options) {
     });
     self.rateLimiterEnabled = options.rateLimiterEnabled;
 
+    self.partialAffinityEnabled = options.partialAffinityEnabled;
+    self.minPeersPerWorker = options.minPeersPerWorker || DEFAULT_MIN_PEERS_PER_WORKER;
+    self.minPeersPerRelay = options.minPeersPerRelay || DEFAULT_MIN_PEERS_PER_RELAY;
+
     self.egressNodes.on('membershipChanged', onMembershipChanged);
```
Raynos: Can refresh on membershipChanged ?

test/partial-affinity.js
kriskowal: The trouble is that we are not yet reaping unneeded peers, so we’ll expect any number of connections between partially connected and fully connected. The only case we can assert is invalid is “not connected”, for which there is ample evidence if the requests fail.

service-proxy.js
kriskowal: Debatable whether that is useful. The affinity gets updated upon every advertise and these happen with some regularity. Doing the refresh on membership changes would only make the shift in traffic more sudden.

Raynos: #legit

service-proxy.js
```diff
@@ -321,11 +328,100 @@ ServiceDispatchHandler.prototype.refreshServicePeer =
...
+            workers.length,
+            Math.max(
+                self.minPeersPerRelay,
+                self.minPeersPerWorker * ratio
+            )
+        )
+    );
+    var stop = (start + length) % workers.length;
+
+    self.logger.info('Refreshing service peer affinity', self.extendLogInfo({
```
Raynos: This log is O(n) advertise. 

Once we've proven the feature with confidence we don't want this.

We may want a log for "new peers" or "removing peers" though.

Hyperbahn should be silent under steady state. We can keep this here for now.

jcorbin:
rebase and then let's merge this for a 2.11cut

Raynos:
lgtm.

PR #7, opened 2015-09-24
TimeSeries test to verify retries handle a single Busy worker.

Raynos:
(This PR is rebased from the old repository).

This test demonstrates that retries can make a single busy
Hyperbahn worker un-noticable.

r: @jcorbin @kriskowal @rf
cc: @anson627


jcorbin:
should probably rebase on latest master (post 10.6) and then we can merge?

jcorbin:
I did the rebaes

PR #8, opened 2015-09-24
larch: level routing backend; debugs skip reservoir

rf:
r: @Raynos @jcorbin 

Adds a new Larch backend called `LevelRouterBackend` which routes to a set of Larch backends based on level. Debugs will skip the reservoir and go straight to Logtron to be written to disk; all other logs will be sent to the reservoir to be sampled, then sent to Logtron.

### DIFF
--- a/lib/larch/base-backend.js
+++ b/lib/larch/base-backend.js
@@ -59,15 +59,11 @@ function BaseBackend(options) {
 BaseBackend.prototype.log = function log(record, cb) {};
 
 BaseBackend.prototype.bootstrap = function bootstrap(cb) {
-    if (typeof cb === 'function') {
-        cb();
-    }
+    cb();
 };
 
-BaseBackend.prototype.destroy = function bootstrap(cb) {
-    if (typeof cb === 'function') {
-        cb();
-    }
+BaseBackend.prototype.destroy = function destroy(cb) {
+    cb();
 };
 
 BaseBackend.prototype.logMany = function logMany(records, cb) {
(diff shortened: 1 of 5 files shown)


Raynos:
Please drop all debugs on the floor. #donotwant

Raynos:
This approach is too complicated and generic. 

Let's just implement `debug` method in larch to do a no-op and get rid of level-router and drop backend.

lib/larch/base-backend.js
```diff
@@ -59,11 +59,15 @@ function BaseBackend(options) {
 BaseBackend.prototype.log = function log(record, cb) {};
 
 BaseBackend.prototype.bootstrap = function bootstrap(cb) {
-    cb();
+    if (typeof cb === 'function') {
```
jcorbin: please don't introduce optional callback pattern, inject noop if you don't care.

lib/larch/base-backend.js
```diff
@@ -76,6 +76,8 @@ BaseBackend.prototype.logMany = function logMany(records, cb) {
     }
 
     function logsDone(ignored, results) {
-        cb(Errors.resultArrayToError(results, 'larch.log-many.many-errors'));
+        if (typeof cb === 'function') {
```
jcorbin: shouldn't be optional

lib/larch/drop-backend.js
```diff
@@ -0,0 +1,51 @@
...
+
+    var self = this;
+
+    BaseBackend.call(self);
+}
+
+util.inherits(DropBackend, BaseBackend);
+
+DropBackend.prototype.log = function log(record, cb) {
+    if (typeof cb === 'function') {
```
jcorbin: not optional

lib/larch/drop-backend.js
```diff
@@ -0,0 +1,51 @@
...
+        cb();
+    }
+};
+
+DropBackend.prototype.logMany = function logMany(records, cb) {
+    if (typeof cb === 'function') {
```
jcorbin: not optional

jcorbin:
lgtm

PR #20, opened 2015-09-30
First pass at README

Raynos:
We can improve on this some more!

r: @mranney @jcorbin


mranney:
This looks great. I vote that we merge this and iterate on it.

PR #23, opened 2015-10-02
Stabilize unadvertise test part two

anson627:
This is hyperbahn part of change to stabilize unadvertise test.

unadvertise event is not sufficient to confirm unadvertisement success on both server and client sides, so wait for all in connections to be removed.

This change depends on a tchannel change to be published and linked:
https://github.com/uber/tchannel-node/pull/33

Multiple test runs passed locally
```bash
for i in {1..50}; do node unadvertise.js; done
```

r: @Raynos @ShanniLi

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "safe-json-parse": "4.0.0",
     "static-config": "2.1.0",
     "tape-cluster": "2.0.1",
-    "tchannel": "3.2.3",
+    "tchannel": "3.3.1",
     "thriftify": "1.0.0-alpha14",
     "uber-statsd-client": "1.3.2",
     "uncaught-exception": "5.0.0",
--- a/test/hyperbahn-client/unadvertise.js
+++ b/test/hyperbahn-client/unadvertise.js
@@ -58,6 +58,7 @@ function runTests(HyperbahnCluster) {
         steveHyperbahnClient.advertise();
 
         function onAdvertised() {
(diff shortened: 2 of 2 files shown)


Raynos:
lgtm.

ShanniLi:
:+1:

anson627:
update tchannel to 3.3.1, tests are passing, so check-in

jcorbin:
Reverting this, we were not ready to move to tchannel 3.3 yet

package.json
```diff
@@ -41,7 +41,7 @@
     "safe-json-parse": "4.0.0",
     "static-config": "2.1.0",
     "tape-cluster": "2.0.1",
-    "tchannel": "3.2.3",
+    "tchannel": "3.3.1",
```
jcorbin: Why even did a test change upgrade tchannel in passing? Please don't just slip in tchannel upgrades without explicit approval, especially after initial approval like happened on this PR.

anson627: this test change is depend on a tchannel change:
https://github.com/uber/tchannel-node/pull/33/files#diff-e3ba5987085d79033122a81a2ddc357cR385

the tchannel change exposed an extra parameter consumed by the test.

without the tchannel change being upgraded, the test uses some parameter which doesn't exist, and therefore fails.

anson627:
ok, I will merge master and re-send the PR.

Issue #17, opened 2015-09-26
Rate-limiter should drop requests when the service RPS is larger than 2x of the limit


ShanniLi:
Otherwise, Hyperbahn nodes can still be stressed out by responding with error frames.

One thing though, 2x may not be aggressive enough. E.g., if the current rate-limit is 1200, 2x would be 2400, the way out of a node's bandwidth. We may want to use the total limit per node as the threshold for dropping requests.  

cc @Raynos @jcorbin @kriskowal

Raynos:
@ShanniLi 

Math.max(totalRPS, 2* rate) is +1

kriskowal:
We should also consider training the circuit breaker to drop requests silently over a certain rate of declines.

ShanniLi:
@kriskowal The reason to drop requests is that Hyperbahn nodes cannot handle current RPS, i.e., too busy to even return busy frames. I think circuit breaker is fine in this scenario since rate limiter should be enough to handle it.

PR #24, opened 2015-10-03
kill switch a service if it is too much over the rate limit

ShanniLi:
r @Raynos @kriskowal @jcorbin 

This change should protect Hyperbahn node from CPU spikes, etc.
ref: #17

### DIFF
--- a/rate_limiter.js
+++ b/rate_limiter.js
@@ -282,7 +282,7 @@ function updateServiceLimit(serviceName, limit) {
     // update ks counter
     counter = self.ksCounters[serviceName];
     if (counter) {
-        counter.rpsLimit = self.ksLimitForService(serviceName);
+        counter.rpsLimit = self.killSwitchLimitForService(serviceName);
     }
 };
 
@@ -296,7 +296,7 @@ function updateTotalLimit(limit) {
     var serviceNames = Object.keys(self.ksCounters);
     for (var i = 0; i < serviceNames.length; i++) {
         var counter = self.ksCounters[serviceNames[i]];
-        counter.rpsLimit = self.ksLimitForService(serviceNames[i]);
+        counter.rpsLimit = self.killSwitchLimitForService(serviceNames[i]);
     }
 };
 
@@ -329,8 +329,8 @@ function createServiceCounter(serviceName) {
     return counter;
 };
 
-RateLimiter.prototype.ksLimitForService =
-function ksLimitForService(serviceName) {
+RateLimiter.prototype.killSwitchLimitForService =
+function killSwitchLimitForService(serviceName) {
     var self = this;
     // allow total RPS limit to kick in first
     var rpsLimit = self.totalRpsLimit + 5;
@@ -357,7 +357,7 @@ function createKillSwitchCounter(serviceName) {
     if (!counter) {
         counter = RateLimiterCounter({
             numOfBuckets: self.numOfBuckets,
-            rpsLimit: self.ksLimitForService(serviceName)
+            rpsLimit: self.killSwitchLimitForService(serviceName)
         });
         self.ksCounters[serviceName] = counter;
     }
--- a/rate_limiter.js
+++ b/rate_limiter.js
@@ -143,6 +143,21 @@ function refreshCounter(counter, rpsStatsName, rpsLimitStatsName, createStatsTag
     counter.refresh();
 };
 
+RateLimiter.prototype.refreshEachCounter =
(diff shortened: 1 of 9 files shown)


rate_limiter.js
```diff
@@ -294,6 +329,41 @@ function createServiceCounter(serviceName) {
     return counter;
 };
 
+RateLimiter.prototype.ksLimitForService =
```
kriskowal: please consistently spell out acronyms, ks/killSwitch

rate_limiter.js
```diff
@@ -294,6 +329,41 @@ function createServiceCounter(serviceName) {
...
 };
 
+RateLimiter.prototype.killSwitchLimitForService =
+function killSwitchLimitForService(serviceName) {
+    var self = this;
+    // allow total RPS limit to kick in first
+    var rpsLimit = self.totalRpsLimit + 5;
+    if (self.serviceCounters[serviceName] &&
+        self.serviceCounters[serviceName].rpsLimit * 2 < rpsLimit) {
+        rpsLimit = self.serviceCounters[serviceName].rpsLimit * 2;
```
kriskowal: Please lift the “rate limiter kill / drop factor” of 2 to a constant at the top of the file.

service-proxy.js
```diff
@@ -172,6 +172,24 @@ function rateLimit(req, buildRes) {
     var isExitNode = self.isExitFor(req.serviceName);
     if (isExitNode) {
         self.rateLimiter.createServiceCounter(req.serviceName);
+        self.rateLimiter.createKillSwitchCounter(req.serviceName);
```
kriskowal: Perhaps this feature should be at ingress instead of egress, or on both sides.

service-proxy.js
```diff
@@ -172,6 +172,24 @@ function rateLimit(req, buildRes) {
...
+    }
+
+    // apply kill switch safe guard first
+    if (isExitNode && self.rateLimiter.shouldKillSwitchService(req.serviceName)) {
```
kriskowal: Consider calling this: `rateLimiter.shouldDrop`. This feature is evidently completely distinct from the kill switch.

kriskowal:
This looks like a good feature, nearly ready for production validation. Are you confident that existing tests adequately verify that this feature does not interfere with traffic beneath the rate limit * drop factor? Perhaps it would be good to write a test to verify that case.

service-proxy.js
Raynos: lets punt on "should be at ingress" until we have rolled out the favor incoming and drop outgoing socket in all languages.

service-proxy.js
Raynos: we REALLY want this feature for total request.

Having a kill switch kick in when 2 * total limit would be great for defending co-tenant workers.

ShanniLi: @kriskowal The reason I use kill-switch is that it has the same behavior as kill-switch. For further steps, we may extend this feature more to be combined with kill-switch.

kriskowal: I’m super glad this isn’t coupled to the kill switch proper. There would have been complications with rate limit kill kill switches overriding manual kill switches. I presume that this kill switch lets traffic through when the rate drops to a reasonable level (no test for that case).

ShanniLi: Good point about the test! I will add it. W.r.t. the name, I would like to use kill switch, since they are essentially doing the same thing: through manual or through rate limiting. :)

kriskowal: I am not going to block on a name change.

ShanniLi:
@kriskowal I have unit test to validate that traffic should not be affected when it is under the limit. I have also written a stress (integration) test for that.

ShanniLi:
@Raynos @kriskowal Fixed most comments. Will do kill switch for total limit in next commit.

rate_limiter.js
```diff
@@ -107,70 +113,77 @@ function RateLimiter(options) {
...
-                'gauge',
-                counter.rpsLimit,
-                statsTag
-            ));
-        }
-        counter.refresh();
+    serviceNames = Object.keys(self.ksCounters);
+    for (i = 0; i < serviceNames.length; i++) {
+        counter = self.ksCounters[serviceNames[i]];
+        self.refreshCounter(counter,
```
jcorbin: looks to me like an opportunity to factor out:
```
RateLimiter#refreshEachCounter(statsName, limitStatsName, createTag, counters)
```

ShanniLi: will do

Raynos:
good enough. lets get it in.

PR #26, opened 2015-10-05
Eliminate error log noise around declined advertises

jcorbin:
Towards deploy confidence, uses the new `tchannel/errors.isFatal(err)` predicate from https://github.com/uber/tchannel-node/pull/34.

@kriskowal @Raynos


kriskowal:
:ok:

PR #28, opened 2015-10-05
Remove readable-stream as dependency

anson627:
git grep shows hyperbahn is not using readable-stream anymore, and the version seems to break npm shrinkwrap if anyone has a dependency on hyperbahn.

r: @ShanniLi @Raynos


Raynos:
Get rid of it :D.

lgtm.

PR #30, opened 2015-10-05
Writing a stable test is hard

anson627:
Unfortunately the unadvertise test is still not 100% stable after my last attempt to fix.

The connection removed count can go below 0 twice and cause callback to fire twice.

Have run unadvertise.js 100 times in a loop without failing.

r: @Raynos


Raynos:
I've got an alternative and cleaner fix on a branch.

I'll try and get that in within an hour.

anson627:
@Raynos sure, close this.

PR #32, opened 2015-10-07
make ringpop more resilient to DDOSing itself

Raynos:
Turns out that pingReq is expensive.

Turns out having a higher ping latency gets us more 9999s

r: @jcorbin @rf

cc: @jwolski @danielheller


jcorbin:
lgtm

PR #33, opened 2015-10-08
Service proxy periodic gauges

kriskowal:


kriskowal:
r @Raynos @jcorbin @rf

Raynos:
Please run hyperbahn benchmarks and look for regressions.

Otherwise lgtm.

service-proxy.js
```diff
@@ -626,6 +636,61 @@ function isExitFor(serviceName) {
...
+ServiceDispatchHandler.prototype.emitPeriodicStats =
+function emitPeriodicStats() {
+    var self = this;
+    self.periodicStatsHandle = null;
+
+    var serviceNames = Object.keys(self.channel.subChannels);
+    for (var index = 0; index < serviceNames.length; index++) {
+        var serviceName = serviceNames[index];
+        var serviceChannel = self.channels.subChannels[serviceName];
+        self.emitPeriodicServiceStats(serviceChannel, serviceName);
```
jcorbin: pretty sure `serviceChannel.serviceName === serviceName`, could get away without the extra arg

service-proxy.js
```diff
@@ -626,6 +636,61 @@ function isExitFor(serviceName) {
...
+            var connection = peer.connections[cIndex];
+            if (connection.direction === 'in') {
+                incoming++;
+            } else if (connection.direction === 'out') {
+                outgoing++;
+            }
+        }
+    }
+
+    self.statsd.gauge(prefix + 'peers', hostPorts.length);
```
jcorbin: these strings look ripe for caching to me... #futureworkthomaybe

service-proxy.js
```diff
@@ -626,6 +636,61 @@ function isExitFor(serviceName) {
...
+    self.statsd.gauge(prefix + 'connections.any', anyway);
+};
+
+ServiceDispatchHandler.prototype.requestPeriodicStats =
+function requestPeriodicStats() {
+    var self = this;
+    if (self.periodicStatsHandle) {
+        return;
+    }
+    self.periodicStatsHandle = self.channel.timers.setTimeout(self.boundEmitPeriodicStats, self.statsPeriod);
```
jcorbin: wrap this in a `if (!self.channel.destroyed)` so that we don't hang process restarts

service-proxy.js
kriskowal: #future sounds good to me. Since these are gauges, I’ve deliberately dropped the interval to 30s to keep it cool.

kriskowal:
I’ve posted an update that ensures the timer gets cancelled when the service proxy is destroyed, and a destroyed flag to prevent it from the interval loop from being restarted. I’ve also made the constructor emit stats immediately so that the code is covered in tests—otherwise the tests and benchmarks do not run long enough for the first interval to conclude. I would like to defer the remaining feedback.

PR #35, opened 2015-10-08
Improved hyperbahn local starter script. 

malandrew:
Use window names instead of window indices.
Write ~/.hyperbahn-hosts.json file and a ~/tcurlrc file pointing to that hostlist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,14 @@ To use a service over Hyperbahn:
 
  - `git clone git@github.com:uber/hyperbahn`
  - `cd hyperbahn`
- - `npm install`
- - `npm install tcurl --global`
  - `./hyperbahn-dev.sh`
 
 This will spawn a 3 window tmux session with a two-node hyperbahn
-cluster running. The third window runs the health checks of both
-nodes to double check that the hyperbahn cluster is healthy.
+cluster running. It will also run `npm install` if necessary and install
+tcurl globally if possible and fallback to local installation if the npm
+global folder is not writeable. The third window runs the health checks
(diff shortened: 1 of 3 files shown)


Raynos:
#toomuchcode

#donotwant

Raynos:
Really tidy. lgtm.

hyperbahn-dev.sh
```diff
@@ -1,14 +1,24 @@
 #!/bin/sh 
 set -e
 
+H0="127.0.0.1:21300"
+H1="127.0.0.1:21301"
+
+if [ ! -d ./node_modules ]; then
```
Raynos: Please `which tcurl` and `npm install tcurl --global` if not tcurl.

Please also update README

Raynos:
Nice. Ship it.

README.md
```diff
@@ -53,13 +53,13 @@ To use a service over Hyperbahn:
...
- - `npm install tcurl --global`
  - `./hyperbahn-dev.sh`
 
 This will spawn a 3 window tmux session with a two-node hyperbahn
-cluster running. The third window runs the health checks of both
-nodes to double check that the hyperbahn cluster is healthy.
+cluster running. It will also run `npm install` if necessary and install
+tcurl globally if possible and fallback to local installation if the npm
+global folder is not writeable. The third window runs the health checks of both nodes using tcurl to double check that the hyperbahn cluster is
+healthy.
```
prashantv: nit: why is this on a separate line, but the above line is too long? We should format to match the above line length

malandrew: nice catch. fixed.

Issue #25, opened 2015-10-05
Rate limiter should clear edge counters that have a zero RPS


ShanniLi:
Otherwise, the list is going to keep growing and create an unnecessary overhead.

cc: @Raynos @kriskowal @jcorbin

Raynos:
@ShanniLi great idea :)

Raynos:
@ShanniLi EdgeCounter is a prototype right ? ;)

ShanniLi:
@Raynos live and learn :)

PR #37, opened 2015-10-09
remove edge counters that have a zero qps

ShanniLi:
r @Raynos @kriskowal 

ref: #25


rate_limiter.js
```diff
@@ -155,6 +155,10 @@ function refreshEachCounter(counters, rpsStatsName, rpsLimitStatsName, createSta
             createStatsTag,
             serviceNames[i]
         );
+
+        if (removeZeroRps && counter.rps === 0) {
```
Raynos: legit.

kriskowal:
twas brillig and greenstamp

Issue #41, opened 2015-10-09
Audit requests for assumed existence of req.connection


kriskowal:
We have observed uncaught exceptions in production due to the assumption that requests have a connection. Specifically in ServiceProxy rate limiting, we need to re-evaluate two occasions where this assumption exists.

kriskowal:
sending to @ShanniLi to investigate

PR #42, opened 2015-10-09
bug fix: kill switch should use req.operations instead of req.connection.ops

ShanniLi:
r @kriskowal @Raynos 
ref: #41, #40


ShanniLi:
This should be always on the path:
https://github.com/uber/tchannel-node/blob/master/operations.js#L226

Raynos:
lgtm.

PR #44, opened 2015-10-09
Remove feature for testing circuits for a single service

kriskowal:
This turned out to be unnecessary and we never used it.


kriskowal:
r @Raynos some low-priority house-keeping.

Raynos:
Thanks. lgtm.

jcorbin:
Yup, especially a good drop since we'll need to add back similar-ish code someday to allow for disruption api.

PR #46, opened 2015-10-09
Make relay ad faster

jcorbin:
- stop doing expensive exit nodes lookup for safety
- instead do safety check within service proxy on channel mode

r @Raynos @kriskowal @rf


Raynos:
Tests pass. #cool

Raynos:
lgtm.

PR #47, opened 2015-10-09
remote config for kill-switch

ShanniLi:
r @Raynos @jcorbin  @kriskowal

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -421,14 +421,14 @@ ApplicationClients.prototype.updateKValues = function updateKValues() {
 
 ApplicationClients.prototype.updateKillSwitches = function updateKillSwitches() {
     var self = this;
-    self.serviceProxy.unblockAll();
+    self.serviceProxy.unblockAllRemoteConfig();
     var killSwitches = self.remoteConfig.get('killSwitch', []);
 
     for (var i = 0; i < killSwitches.length; i++) {
         var value = killSwitches[i];
         var edge = value.split('~~');
-        if (edge.length === 2 && (edge[0] !== '*' || edge[1] !== '*')) {
-            self.serviceProxy.block(edge[0], edge[1]);
+        if (edge.length === 2 && value !== '*~~*') {
+            self.serviceProxy.blockRemoteConfig(edge[0], edge[1]);
         }
     }
 };
--- a/clients/index.js
+++ b/clients/index.js
@@ -334,6 +334,7 @@ ApplicationClients.prototype.onRemoteConfigUpdate = function onRemoteConfigUpdat
     self.updateExemptServices();
     self.updateRpsLimitForServiceName();
     self.updateKValues();
+    self.updateKillSwitches();
     self.updateReservoir();
     self.updatePartialAffinityEnabled();
 };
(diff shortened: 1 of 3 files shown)


clients/index.js
```diff
@@ -417,3 +418,17 @@ ApplicationClients.prototype.updateKValues = function updateKValues() {
         self.serviceProxy.updateServiceChannels();
     }
 };
+
+ApplicationClients.prototype.updateKillSwitches = function updateKillSwitches() {
+    var self = this;
+    self.serviceProxy.unblockAll();
+    var killSwtiches = self.remoteConfig.get('killSwitch', []);
```
rf: `killSwitches` (fix mispelling) please

clients/index.js
```diff
@@ -417,3 +418,17 @@ ApplicationClients.prototype.updateKValues = function updateKValues() {
...
+
+    for (var i = 0; i < killSwtiches.length; i++) {
+        var value = killSwtiches[i];
+        var edge = value.split('~~');
+        if (edge.length === 2 && (edge[0] !== '*' || edge[1] !== '*')) {
```
rf: why not allow wildcard killSwitches in flipr?

clients/index.js
```diff
@@ -417,3 +418,17 @@ ApplicationClients.prototype.updateKValues = function updateKValues() {
         self.serviceProxy.updateServiceChannels();
     }
 };
+
+ApplicationClients.prototype.updateKillSwitches = function updateKillSwitches() {
+    var self = this;
+    self.serviceProxy.unblockAll();
```
rf: Won't this remove any manual ones we run? I think it would still be useful to have an immediate fan out kill switch rather than relying on remote-config to load the changes.

ShanniLi: Yes, this will.
Remote config will be the source of truth. So, the steps should be:
1. set up the remote config.
2. run admin command for immediate effect.

clients/index.js
ShanniLi: It wasn't supported in admin tool neither. I guess it doesn't make sense to kill everything.

clients/index.js
ShanniLi: that is a remote config property.

ShanniLi: sorry, misread the comment. Will fix.

clients/index.js
jcorbin: I also don't like this approach of: "unblock all, then block these".

Please instead: "block these, then unblock any that I didn't just block".

Raynos: do not want unblockAll().

Use the build in delta feature in remoteConfig. We actually have to listen to the event and understand current and previous state and do proper deltas.

clients/index.js
```diff
@@ -417,3 +418,17 @@ ApplicationClients.prototype.updateKValues = function updateKValues() {
...
+    var killSwitches = self.remoteConfig.get('killSwitch', []);
+
+    for (var i = 0; i < killSwitches.length; i++) {
+        var value = killSwitches[i];
+        var edge = value.split('~~');
+        if (edge.length === 2 && (edge[0] !== '*' || edge[1] !== '*')) {
```
Raynos: easier to read as `edge.length ===2 && value !== '*~~*'`

Raynos:
Needs more work.

We need to rely on deltas and do things #coolstory

Raynos:
lgtm.

PR #48, opened 2015-10-09
switch for lazy relaying

rf:
fix the flipr flag `lazy.handling.enabled` to also flip lazy relaying as well, and delay turning off lazy handling by 30s when switching off

r @jcorbin @Raynos


jcorbin:
unsure why the test fail, I can't get them to pass locally either

Raynos:
lgtm. @rf shipit

PR #51, opened 2015-10-13
Fix Hyperbahn ThriftRW compatibility tests

kriskowal:
This introduces a few fixes that will let us upgrade ThriftRW in TChannel, currently failing for the link test. This is blocking a fix for schema migration.

1. Fixes a typo so the empty-query test uses the intended IDL file. It was already there.
2. Fixes the corresponding error message.
3. Relaxed the constraint for unrecognized exceptions. They were transport errors previously. Now they are application errors with no body or typeName. In the future, they will be reified ThriftUnrecognizedException instances. The only guarantee going forward is that ok will be false.

r @Raynos

### DIFF
--- a/test/hyperbahn-client/hostports.js
+++ b/test/hyperbahn-client/hostports.js
@@ -215,7 +215,7 @@ function runTests(HyperbahnCluster) {
             serviceName: 'hyperbahn',
             hasNoParent: true
         });
-        var badSource = fs.readFileSync(path.join(__dirname, 'bad-hyperbahn-empty-req-body.thrift'), 'utf8');
+        var badSource = fs.readFileSync(path.join(__dirname, 'bad-hyperbahn-empty-query.thrift'), 'utf8');
         var badThrift = new TChannelAsThrift({source: badSource});
         badThrift.send(request,
             'Hyperbahn::discover',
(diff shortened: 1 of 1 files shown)


test/hyperbahn-client/hostports.js
```diff
@@ -268,18 +269,8 @@ function runTests(HyperbahnCluster) {
...
         function onResponse(err, res) {
-            assert.ok(err, 'should be error');
-            assert.equals(err.type, 'tchannel-thrift-handler.parse-error.body-failed',
-                'error should be parser error');
-            assert.ok(err.message.indexOf(
-                'Could not parse body (arg3) argument.\n' +
-                'Expected Thrift encoded arg3 for endpoint Hyperbahn::discover.') === 0,
-                'error message should be a parsing failure');
-
-            var items = cluster.logger.items();
```
Raynos: Do not remove the logger please :)

test/hyperbahn-client/hostports.js
```diff
@@ -268,18 +269,8 @@ function runTests(HyperbahnCluster) {
...
-            assert.ok(items.length > 0 && items[0].msg === 'Got unexpected invalid thrift for arg3',
-                'Do not miss the error log');
-
+            // TODO expect UnrecognizedException
+            assert.ok(err || !res.ok, 'should be error (possibly unrecognized application error)');
```
Raynos: let's just improve these asserts and change them for `thriftrw`

kriskowal: we will have to revisit when unrecognized exceptions are implemented. atm, you get a pretty bad middle ground.

Raynos:
good enough; move along :)

PR #55, opened 2015-10-14
service-proxy: fix calls to sendLazyErrorFrameForReq

rf:
Fixes calls to the method I refactored in https://github.com/uber/tchannel-node/pull/78

r @Raynos @jcorbin


Raynos:
Also update tchannel ;)

lgtm.

PR #58, opened 2015-10-16
Support spawning multiple workers in a seperate process

Raynos:
This is the WIP for advertise benchmark. The first step
was to get multiple hyperbahn workers in the benchmark
so we can test ringpop / ad / relay-ad etc.

A secondary step would be to update the bench_client
in tchannel-node to send "ad" messages.

### DIFF
--- a/benchmarks/hyperbahn-worker.js
+++ b/benchmarks/hyperbahn-worker.js
@@ -28,6 +28,8 @@ var StaticConfig = require('static-config');
 
 var HyperbahnApplication = require('../app.js');
 
+module.exports = HyperbahnWorker;
+
 function HyperbahnWorker(opts) {
     /*eslint max-statements: [2, 25]*/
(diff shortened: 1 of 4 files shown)


benchmarks/hyperbahn-worker.js
```diff
@@ -107,10 +107,12 @@ HyperbahnWorker.prototype.createConfig = function createConfig() {
                 leafHost: '127.0.0.1',
                 leafPort: self.kafkaPort
             },
+            // 'clients.logtron.console': true,
```
jcorbin: @Raynos this causes logtron to try to write into `/var/log/hyperbahn` which is an `EACCES` in travis.

Raynos: How does turning on console also turn on disk... Wat.

We can change the logFile to be /tmp

jcorbin:
Okay, this now works and will be the basis of future work, so merging.

PR #59, opened 2015-10-16
Add remote config support for configuring the peer reaper

kriskowal:
r @Raynos cc @jcorbin

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -341,6 +341,7 @@ ApplicationClients.prototype.onRemoteConfigUpdate = function onRemoteConfigUpdat
     self.updateRpsLimitForServiceName();
     self.updateKValues();
     self.updateReservoir();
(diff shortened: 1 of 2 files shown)


clients/index.js
```diff
@@ -392,6 +393,14 @@ ApplicationClients.prototype.updateRateLimitingEnabled = function updateRateLimi
     }
 };
 
+ApplicationClients.prototype.updateReapPeersPeriod = function updateReapPeersPeriod() {
+    var self = this;
+    var period = self.remoteConfig.get('peerReaper.period', 0);
+    if (period !== self.serviceProxy.reapPeersPeriod) {
```
Raynos: This gaurd exists inside the method; just always call the method.

Raynos:
lgtm :)

Issue #61, opened 2015-10-16
BUG: When reaping peers we get a null pointer sad.


Raynos:
```
TypeError: Cannot read property 'connections' of null
    at ServiceDispatchHandler.reapPeers (/var/cache/udeploy/r/hyperbahn/sjc1-produ-0000000072-v1/node_modules/hyperbahn/service-proxy.js:819:29)
    at reapPeers [as _onTimeout] (/var/cache/udeploy/r/hyperbahn/sjc1-produ-0000000072-v1/node_modules/hyperbahn/service-proxy.js:110:14)
    at Timer.listOnTimeout [as ontimeout] (timers.js:112:15)
```

PR #62, opened 2015-10-16
clean up book keeping object to avoid null pointers

Raynos:
Fixes: https://github.com/uber/hyperbahn/issues/61

r: @kriskowal @rf @jcorbin

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -832,6 +832,8 @@ function reapPeers() {
         }
 
         self.channel.peers.delete(hostPort);
+        delete self.peersToReap[hostPort];
+        delete self.connectedPeers[hostPort];
     }
 
     self.peersToReap = self.connectedPeers;
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -816,6 +816,10 @@ function reapPeers() {
     for (i = 0; i < peersToReap.length; i++) {
         var hostPort = peersToReap[i];
         var peer = self.channel.peers.get(hostPort);
+        if (!peer) {
+            continue;
+        }
+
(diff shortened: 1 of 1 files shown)


service-proxy.js
```diff
@@ -832,6 +832,8 @@ function reapPeers() {
         }
 
         self.channel.peers.delete(hostPort);
+        delete self.peersToReap[hostPort];
```
kriskowal: this should be unnecessary—peersToReap gets stomped on :839

service-proxy.js
```diff
@@ -832,6 +832,8 @@ function reapPeers() {
...
+        delete self.connectedPeers[hostPort];
```
kriskowal: This should be unhelpful, because the peer cannot be reaped on the subsequent reaping. This introduces a leak.

Raynos: no ? Question mark. We just deleted it; i.e. `peers.delete(hp);`

Raynos:
What this PR does is Once we remove a Peer object from memory we assume it is completely gone; i.e. `peers.delete(hp)` should remove the TChannelPeer object from memory completely.

After doing so I'm cleaning up the objects to remove a reference to this host port.

The only way it can come back is another open connection or `"ad"` i.e. the `refreshServicePeer()` method that get's called when we receive a `"relay-ad"`.

service-proxy.js
Raynos: Yeah I think this is just me being silly.

jcorbin:
lgtm now with less delet

kriskowal:
:ok:

PR #64, opened 2015-10-17
Default to bin/bash for making script work better

Raynos:
r: @lopter @malandrew


lopter:
Use `/usr/bin/env bash` like you would do for any interpreted language, this will not work on (BSD) systems that don't link shells into `/bin`

PR #66, opened 2015-10-17
Add partial affinity peer pruning

Raynos:
I've updated the code here to make it work with
dead peer reaper.

I've also fixed a few bugs here and there.

r: @jcorbin @rf @kriskowal

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -776,7 +776,9 @@ function ensurePartialConnections(chan, serviceName, reason, now) {
     }
 
     var connectedPeers = self.connectedServicePeers[serviceName];
+    var connectedPeerKeys = connectedPeers ? Object.keys(connectedPeers) : [];
     var toConnect = [];
+    var toDisconnect = [];
     var isAffine = {};
     var i;
     var worker;
@@ -792,8 +794,14 @@ function ensurePartialConnections(chan, serviceName, reason, now) {
             toConnect.push(worker);
         }
     }
+    for (i = 0; i < connectedPeerKeys.length; i++) {
+        worker = connectedPeerKeys[i];
+        if (!isAffine[worker] && !self.peersToPrune[worker]) {
+            toDisconnect.push(worker);
+        }
+    }
 
-    if (!toConnect.length) {
+    if (!toConnect.length && !toDisconnect.length) {
         result.noop = true;
         return result;
     }
@@ -802,7 +810,8 @@ function ensurePartialConnections(chan, serviceName, reason, now) {
         serviceName: serviceName,
         reason: reason,
         partialRange: range,
-        toConnect: toConnect
+        toConnect: toConnect,
+        toDisconnect: toDisconnect
     }));
 
     var peer;
@@ -811,9 +820,10 @@ function ensurePartialConnections(chan, serviceName, reason, now) {
         self.ensurePeerConnected(serviceName, peer, 'service peer affinity change', now);
     }
 
-    // TODO Drop peers that no longer have affinity for this service, such
-    // that they may be elligible for having their connections reaped.
-
+    for (i = 0; i < toDisconnect.length; i++) {
+        peer = self._getServicePeer(chan, toDisconnect[i]);
+        self.ensurePeerDisconnected(serviceName, peer, 'service peer affinity change', now);
+    }
     return result;
 };
 
--- a/test/config/with_partial_affinity.json
+++ b/test/config/with_partial_affinity.json
@@ -1,0 +1,3 @@
+{
+    "partialAffinity.enabled": true
+}
--- a/test/forward/dead-remote-reaped.js
+++ b/test/forward/dead-remote-reaped.js
@@ -32,11 +32,16 @@ allocCluster.test('dead exit peers get reaped', {
     size: 10,
     namedRemotes: ['alice', 'alice', 'alice', 'alice', 'alice', 'alice', 'alice'],
     whitelist: [
+        ['info', 'implementing affinity change'],
+        ['info', 'pruning ex-affinity peers'],
+        ['info', 'draining ex-affinity peer'],
+        ['info', 'draining peer'],
         ['info', 'Refreshing service peer affinity'],
         ['info', 'reaping dead peers'],
         ['info', 'reaping dead peer'],
         ['info', 'pruning peers'],
-        ['info', 'draining pruned peer']
+        ['info', 'draining pruned peer'],
+        ['info', 'refreshed peer partially']
     ]
 }, function t(cluster, assert) {
     var activeNum = 3;
--- a/test/forward/dead-remote-reaped.js
+++ b/test/forward/dead-remote-reaped.js
@@ -32,6 +32,10 @@ allocCluster.test('dead exit peers get reaped', {
     size: 10,
     namedRemotes: ['alice', 'alice', 'alice', 'alice', 'alice', 'alice', 'alice'],
     whitelist: [
+        // TODO: this is debt, this should not be an expected log; we should
+        // understand and fix the cause of and need for this audit
+        ['warn', 'partial affinity audit fail'],
+
         ['info', 'implementing affinity change'],
         ['info', 'pruning ex-affinity peers'],
         ['info', 'draining ex-affinity peer'],
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,5 @@ env:
 - HYPERBAHN_REMOTE_TEST_CONFIG=
 - HYPERBAHN_REMOTE_TEST_CONFIG=test/config/with_lazy_relay.json
 - HYPERBAHN_REMOTE_TEST_CONFIG=test/config/with_peer_heap.json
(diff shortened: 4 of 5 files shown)


test/partial-affinity.js
```diff
@@ -103,6 +103,26 @@ allocCluster.test('nominal parameters', {
...
 
+var hundredAlice = [];
+for (var _count = 0; _count < 30; _count++) {
+    hundredAlice.push('alice');
+}
+
+allocCluster.test('big cluster', {
+    size: 30,
+    kValue: 15,
+    namedRemotes: hundredAlice,
```
kriskowal: manyAlices. Specifically 30, but that’s not important right now. https://www.youtube.com/watch?v=VOmD-xqK2Es

service-proxy.js
```diff
@@ -514,38 +520,71 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
     // Open connections to affine peers
     var index;
     var peer;
-    if (start <= stop) { // ... start WITHIN stop ...
-        for (index = start; index < stop; index++) {
-            peer = self.getServicePeer(serviceName, workers[index]);
-            peer.connectTo();
+
+    // fully connected
+    if (range.start === range.stop) {
```
kriskowal: I would say we need to thread modulo through here to distinguish fully connected from not connected at all, but I think we can rule out "not connected at all" on principle. :ok:

test/config/with_partial_affinity.json
```diff
@@ -0,0 +1,3 @@
+{
+    "artialAffinity.enabled": true
```
kriskowal: less artial, more p

service-proxy.js
```diff
@@ -514,38 +520,71 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
+            }
         }
-        for (index = 0; index < stop; index++) {
-            peer = self.getServicePeer(serviceName, workers[index]);
-            peer.connectTo();
+        for (index = range.start; index < range.workers.length; index++) {
+            peer = self.getServicePeer(serviceName, range.workers[index]);
+            if (!peer.isConnected('out')) {
+                peer.connectTo();
+            }
```
kriskowal: This will probably work well enough, but I don’t think it’ll properly reap connections that are still advertising but no longer affine. We may need to follow up by decoupling peer membership (retained by ad) from affine membership (connected by ad if affine, and drain connections when they are no longer affine, which would be triggered by the peer reaping)

service-proxy.js
```diff
@@ -514,38 +520,71 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
+    };
+};
+
+ServiceDispatchHandler.prototype.refreshServicePeerPartially =
+function refreshServicePeerPartially(serviceName, hostPort) {
+    var self = this;
+
+    var range = self.computePartialRange(serviceName, hostPort);
+
+    self.logger.info('Refreshing service peer affinity', self.extendLogInfo({
```
kriskowal: I can’t tell whether the prevailing convention is to capitalize logs or not.

test/forward/dead-remote-peer-partial.js
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        // Reap peers
+        for (i = 0; i < cluster.apps.length; i++) {
+            app = cluster.apps[i];
+            serviceProxy = app.clients.serviceProxy;
+            serviceProxy.reapPeers();
+        }
+
+        // // Some of the peers remain
```
kriskowal: double comment

kriskowal:
End of review. In short, I think this will work, but we will retain connections unintentionally as the worker and relay rings resize.

service-proxy.js
```diff
@@ -497,7 +503,7 @@ function refreshServicePeerPartially(serviceName, hostPort) {
             relays: relays,
             workers: workers
         }));
-        return;
+        return null;
```
jcorbin: why this

service-proxy.js
```diff
@@ -514,38 +520,71 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
+ServiceDispatchHandler.prototype.refreshServicePeerPartially =
+function refreshServicePeerPartially(serviceName, hostPort) {
+    var self = this;
+
+    var range = self.computePartialRange(serviceName, hostPort);
+
+    self.logger.info('Refreshing service peer affinity', self.extendLogInfo({
+        serviceName: serviceName,
+        serviceHostPort: hostPort,
+        relayIndex: range.relayIndex,
```
jcorbin: can we just `range: range` please?

service-proxy.js
```diff
@@ -514,38 +520,71 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
     var peer;
-    if (start <= stop) { // ... start WITHIN stop ...
-        for (index = start; index < stop; index++) {
-            peer = self.getServicePeer(serviceName, workers[index]);
-            peer.connectTo();
+
+    // fully connected
+    if (range.start === range.stop) {
+        for (index = 0; index < range.workers.length; index++) {
+            peer = self.getServicePeer(serviceName, range.workers[index]);
```
jcorbin: okay, this complexity is too much at this point... factor function out imo

service-proxy.js
```diff
@@ -514,38 +520,57 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
+    }
+
+    function connectToRange(lo, hi) {
+        for (var i = lo; i < hi; i++) {
+            var peer = self.getServicePeer(serviceName, range.workers[i]);
+            if (!peer.isConnected('out')) {
+                peer.connectTo();
+            }
+        }
+    }
```
kriskowal: Unnecessary closure

jcorbin: closure is only used down stack, and this is a cold path... but if you must...

jcorbin:
Now ready for review, please feedback on the marked TODOs.

Normal tests pass, working through failures under `HYPERBAHN_REMOTE_TEST_CONFIG=test/config/with_partial_affinity.json` now.

service-proxy.js
```diff
@@ -51,6 +52,9 @@ var RATE_LIMIT_KILLSWITCH = 'killswitch';
 
 var CN_HEADER_BUFFER = new Buffer('cn');
 
+// TODO: ring change trigger
```
Raynos: We need this before going global, we can test on canaries without ringpop change trigger.

service-proxy.js
```diff
@@ -655,8 +730,23 @@ function refreshServicePeerPartially(serviceName, hostPort, now) {
...
+    deleteIndexEntry(self.connectedServicePeers, serviceName, peer.hostPort);
+    deleteIndexEntry(self.connectedPeerServices, peer.hostPort, serviceName);
+
+    var peerServices = self.connectedPeerServices[peer.hostPort];
+    if (!peerServices || isObjectEmpty(peerServices)) {
+        self.peersToPrune[peer.hostPort] = now;
+    }
 };
 
 ServiceDispatchHandler.prototype.removeServicePeer =
```
Raynos: Should we remove indexes here ? this gets called in unadvertise.

Raynos:
This pruning is non-trivial.

I think we need to go write some feature tests.

Something like:

 - Cluster of K
 - 3 named remotes.
 - foo advertises 2 workers;
 - foo advertises another worker after n ttl.
 - foo advertises another worker after n*2 ttl

or

 - Cluster of K
 - 3 named remotes.
 - foo advertises 5 workers;
 - foo does a rolling restart with n ttl between each worker

The kind of asserts we care about are:

 - What are all the open connections on hyperbahn affinity & named remotes
 - How many connections were closed / recycled
 - What is the connection churn.

test/forward/dead-remote-reaped.js
```diff
@@ -141,6 +166,23 @@ function reapClusterPears(cluster, assert, callback) {
     );
 }
 
+function pruneClusterPears(cluster, assert, callback) {
```
Raynos: http://weknowyourdreams.com/image.php?pic=/images/pear/pear-05.jpg

Raynos:
yay. lgtm.

test/forward/dead-remote-reaped.js
kriskowal: Still :pear: :pear: :pear:

kriskowal: I believe you may have mis-spelled "harvest".

service-proxy.js
```diff
@@ -758,8 +760,14 @@ function ensurePartialConnections(chan, serviceName, reason, now) {
             toConnect.push(worker);
         }
     }
+    for (i = 0; i < connectedPeerKeys.length; i++) {
+        worker = connectedPeerKeys[i];
+        if (!isAffine[worker] && !self.peersToPrune[worker]) {
```
kriskowal: :sealofapproval:

kriskowal:
:ok:

jcorbin:
Rebased on (hopefully) less flappy master, pending travis for merge.

PR #67, opened 2015-10-17
Fix peer reaping timer

kriskowal:
r @jcorbin @Raynos

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -827,13 +827,14 @@ function requestReapPeers() {
 ServiceDispatchHandler.prototype.reapPeers =
 function reapPeers() {
     var self = this;
+    self.reapPeersTimer = null;
 
     var i;
     var j;
 
     var peersToReap = Object.keys(self.peersToReap);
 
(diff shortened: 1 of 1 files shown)


service-proxy.js
```diff
@@ -827,13 +827,14 @@ function requestReapPeers() {
 ServiceDispatchHandler.prototype.reapPeers =
 function reapPeers() {
     var self = this;
+    self.reapPeersTimer = null;
```
kriskowal: This is needed such that the next requestReapPeers goes through after this is triggered by a timeout.

service-proxy.js
```diff
@@ -827,13 +827,14 @@ function requestReapPeers() {
...
 
     var i;
     var j;
 
     var peersToReap = Object.keys(self.peersToReap);
 
-    self.logger.info('Reaping dead peers', self.extendLogInfo({
+    self.logger.info('reaping dead peers', self.extendLogInfo({
```
kriskowal: The other infos were making fun of it.

jcorbin:
lgtm, I'm going to ship this now

PR #70, opened 2015-10-19
test-cluster now works with thrift

jskelcy:
If you pass in a thrift file associated with your named remote you can now register with an "as thrift" sub-channel. 

Fun fact: the api is weird so I dont break back-compat.

### DIFF
--- a/test/forward/thrift.js
+++ b/test/forward/thrift.js
@@ -59,7 +59,7 @@ allocCluster.test('register and forward with thrift', {
 
     echoServerRemote.thriftServer.register(
         echoServerRemote.serverChannel, 'echo::thrift_echo', {}, echo
-    )
+    );
 
     function echo(ctx, req, arg2, arg3, cb) {
         cb(null, {
@@ -78,13 +78,13 @@ allocCluster.test('register and forward with thrift', {
     }, onForwarded);
 
     echoServerRemote.thriftClient.request({
-        serviceName: 'echoRemote',
+        serviceName: 'echoRemote'
     }).send('echo::thrift_echo', null, {
         foo: {
             bar: 2,
             baz: 'hi'
         }
-    }, onForwarded)
+    }, onForwarded);
 
     function onForwarded(err, res) {
         assert.ifError(err);
--- a/test/forward/thrift.js
+++ b/test/forward/thrift.js
@@ -31,7 +31,7 @@ var allocCluster = require('../lib/test-cluster.js');
 allocCluster.test('register and forward with thrift', {
     size: 5,
     namedRemotes: ['echoRemote'],
-    remoteSpecs: {
+    remotes: {
         echoRemote: {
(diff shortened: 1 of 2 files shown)


test/forward/thrift.js
```diff
@@ -29,9 +29,14 @@ var someSpec = fs.readFileSync(
 var allocCluster = require('../lib/test-cluster.js');
 
 allocCluster.test('register and forward with thrift', {
-    size: 5
+    size: 5,
+    namedRemotes: ['echoRemote'],
+    remoteSpecs: {
```
Raynos: Let's do some future proofing.

```js
{
    remotes: {
        echoRemote: {
            thriftSpec: path.join(...)
        }
    }
}
```

test/lib/test-cluster.js
```diff
@@ -244,27 +255,49 @@ TestCluster.prototype.createRemote = function createRemote(opts, cb) {
...
+
+    var remote = {
+        clientChannel: clientChannel,
+        serverChannel: serverChannel,
         channel: channel,
         serviceName: opts.serviceName,
         hostPort: null,
-        destroy: destroy
+        destroy: destroy,
+        thriftServerChannel: thriftServerChannel,
```
Raynos: `thriftServer`

test/lib/test-cluster.js
```diff
@@ -244,27 +255,49 @@ TestCluster.prototype.createRemote = function createRemote(opts, cb) {
...
+        thriftClientChannel: thriftClientChannel
```
Raynos: `thriftClient`

test/forward/thrift.js
```diff
@@ -64,6 +75,15 @@ allocCluster.test('register and forward with thrift', {
         }
     }, onForwarded);
 
+    echoServerRemote.thriftClientChannel.request({
```
Raynos: `thriftClient.request({ ... })`

test/forward/thrift.js
```diff
@@ -48,6 +55,10 @@ allocCluster.test('register and forward with thrift', {
         steve.serverChannel, 'echo::thrift_echo', {}, echo
     );
 
+    echoServerRemote.thriftServerChannel.register(
```
Raynos: `thriftServer.register(...)`

test/forward/thrift.js
jskelcy: I feel that. But I think it would be wise to keep the top level property called remoteSpecs. What do you think?

Raynos: Nope; remotes. I want to add other thigns to it like amount and other things so we can deprecate the jank array.

jskelcy:
@Raynos?

Raynos:
I approve.

Raynos:
travis is being a pain. lgtm; lets merge

PR #74, opened 2015-10-20
test-cluster: Use logger.error() instead for better debugger

Raynos:
Emitting an error prints a frustratingly useless stack. Logger.error()
does the right thing (TM)

r: @jcorbin @rf @kriskowal

### DIFF
--- a/test/lib/test-cluster.js
+++ b/test/lib/test-cluster.js
@@ -331,7 +331,10 @@ TestCluster.prototype.createRemote = function createRemote(opts, cb) {
 
     function onRegister(err) {
         if (err) {
-            return self.emit('error', err);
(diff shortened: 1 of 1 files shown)


test/lib/test-cluster.js
```diff
@@ -331,7 +331,10 @@ TestCluster.prototype.createRemote = function createRemote(opts, cb) {
 
     function onRegister(err) {
         if (err) {
-            return self.emit('error', err);
```
jcorbin: sounds like we need to hookup some error events to logger errors then?

jcorbin:
can live with it, merge if you want

test/lib/test-cluster.js
Raynos: screw error events !

PR #76, opened 2015-10-21
Remove npm from engines because its a waste.

Raynos:
It causes confusing warnings in an npm install.

r: @jcorbin @rf


rf:
:+1: why travis tho

PR #77, opened 2015-10-21
move larch into package

rf:
@Raynos @jcorbin 

Larch is in its own repo and published. This removes larch from Hyperbahn and uses Larch out of its package.


Raynos:
lgtm.

PR #84, opened 2015-10-23
Add write buffer mode config

jcorbin:
r @Raynos


Raynos:
please also update the travis.

PR #85, opened 2015-10-24
Improve peer indexing

jcorbin:
This sets the stage to add more indices towards #66:
- factor out general `{add,delete}IndexEntry` utilities
- factor out `{add,delete}PeerIndex` methods
- delete index entries in `reapSinglePeer` to protect against double reap edge cases

r @kriskowal @Raynos

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -486,6 +486,13 @@ function addPeerIndex(serviceName, hostPort) {
     addIndexEntry(self.knownPeers, hostPort, serviceName, true);
 };
 
+ServiceDispatchHandler.prototype.deletePeerIndex =
+function deletePeerIndex(serviceName, hostPort) {
+    var self = this;
+
+    deleteIndexEntry(self.knownPeers, hostPort, serviceName);
+};
+
 ServiceDispatchHandler.prototype.computePartialRange =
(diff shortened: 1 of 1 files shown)


jcorbin:
Interesting, so the commit that breaks test is https://github.com/uber/hyperbahn/commit/5490c36e8c0df1ae72cf5bd7fdce6d5b1d169a57, which I think is more correct behavior, but I need to investigate more as to why `test/forward/dead-remote-reaped.js` fails with it.

service-proxy.js
```diff
@@ -1149,3 +1155,31 @@ function extendLogInfo(info) {
...
+}
+
+/* eslint-disable guard-for-in, no-unused-vars */
+function isObjectEmpty(obj) {
+    for (var prop in obj) {
+        return false;
+    }
+    return true;
+}
+/* eslint-enable guard-for-in, no-unused-vars */
```
Raynos: Soon line 1200. SOON. Then 1500, then 2250, then finally 6000.

jcorbin: Couldn't agree more... this whole file needs to get re-orged and has grown too much... eventually...

Raynos:
lgtm.

PR #86, opened 2015-10-26
ServiceDispatchHandler: downgrade "request missing cn header"

jcorbin:
Downgrade to a warn, add bit more context to log.

r @Raynos @rf


Raynos:
Can we start a new pattern?

If socketRemoteAddr is in ringpop servers/membership then isHyperbahn: true

Probably in conn.extendLogInfo.

Otherwise Lgtm

PR #88, opened 2015-10-26
Improve and correct peer drain

jcorbin:
- clear any prior timer (just in case)
- use `peer.drain` with a timeout
- actually close drained connections
- don't delete the peer until after the drain and close is done

r @Raynos @rf

### DIFF
--- a/app.js
+++ b/app.js
@@ -32,6 +32,7 @@ var ApplicationClients = require('./clients/');
 var ExitNode = require('./exit');
 var EntryNode = require('./entry');
 
+// our call SLA is 30 seconds currently
 var DRAIN_DEADLINE_TIMEOUT = 30 * 1000;
 
 var ApplicationClientsFailureError = WrappedError({
--- a/app.js
+++ b/app.js
@@ -32,9 +32,6 @@ var ApplicationClients = require('./clients/');
 var ExitNode = require('./exit');
 var EntryNode = require('./entry');
 
-// our call SLA is 30 seconds currently
-var DRAIN_DEADLINE_TIMEOUT = 30 * 1000;
-
 var ApplicationClientsFailureError = WrappedError({
     type: 'autobahn.app-clients-failed',
     message: 'Application createClients failed: {origMessage}'
@@ -182,7 +179,8 @@ function startDrain() {
     self.logger.info('got SIGTERM, draining application', self.extendLogInfo({}));
     self.tchannel.drain('shutting down due to SIGTERM', drainedThenClose);
     self.drainDeadlineTimer = self.tchannel.timers.setTimeout(
-        deadlineTimedOut, DRAIN_DEADLINE_TIMEOUT);
+        deadlineTimedOut,
+        self.clients.serviceProxy.drainTimeout);
 
     function drainedThenClose() {
         self.drainedThenClose();
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -939,8 +939,8 @@ function reapSinglePeer(hostPort) {
     }
 
     var serviceNames = Object.keys(self.peersToReap[hostPort]);
-    for (var j = 0; j < serviceNames.length; j++) {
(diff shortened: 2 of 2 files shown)


service-proxy.js
```diff
@@ -950,7 +949,44 @@ function reapSinglePeer(hostPort) {
         self.deletePeerIndex(serviceName, hostPort);
     }
 
-    self.channel.peers.delete(hostPort);
+    // TODO: info log/stat
+
+    peer.drain({
+        reason: 'reaped for expired advertisement',
+        direction: 'both',
```
Raynos: I think we only want to drain 'out'.

There is a race here where we reject incoming connections on a peer that is being reaped but decides to come back.

Raynos: We may want a direction: 'both' but onlyAccept: true semantics.

service-proxy.js
```diff
@@ -950,7 +949,44 @@ function reapSinglePeer(hostPort) {
...
+        peer.close(thenDeleteIt);
+    }
+
+    function thenDeleteIt(err) {
+        if (err) {
+            self.logger.warn('error closing reaped peer, deleting it anyhow', self.extendLogInfo({
+                error: err
+            }));
+        }
+        self.channel.peers.delete(hostPort);
```
Raynos: There are kinds of annoying races that can happen.

We need to fix:

 - https://github.com/uber/tchannel-node/blob/master/connection.js#L477
 - https://github.com/uber/tchannel-node/blob/master/peer.js#L493

It's probably fine but could be "nicer" tm.

Raynos:
:+1: this is an improvement.

The fact that we forgot to close the peer was an important oversight.

service-proxy.js
```diff
@@ -41,6 +41,7 @@ var DEFAULT_MIN_PEERS_PER_WORKER = 5;
 var DEFAULT_MIN_PEERS_PER_RELAY = 5;
 var DEFAULT_STATS_PERIOD = 30 * 1000; // every 30 seconds
 var DEFAULT_REAP_PEERS_PERIOD = 0; // never
+var DEFAULT_DRAIN_TIMEOUT = 30;
```
Raynos: 30 milli seconds :p

service-proxy.js
jcorbin: wat is `onlyAccept: true` semantics?

Also prior was effectively `direction: 'both'` since it just did its own iteration of `peer.connections`... but sure if you want to change even more semantics in this diff, I'll do so ;-)

jcorbin:
Agreed there may be more work to do, but going to just merge this as progress for now.

service-proxy.js
jcorbin: oh good catch... will fix

Raynos:
@jcorbin please 30000 instead of 30.

service-proxy.js
Raynos: Currently there is a "bug" in connection.js

 - on init response flushed
 - get peer for `hostPort`
 - Ignore the fact that the peer might be draining or closing
 - add the incoming connection to the peer
 - #leak 
 - profit

{ direction: 'both', onlyAccept: true } semantics would be something like "drain all connections right now; but if there are any incoming connections during the drain timeout add them to the peer"

We would also need to do a "peer.hasConnections()" in the drain callback check to only shutdown the peer if there were no new incoming connections in the 30 second window

#compexity #:(

Raynos:
Please ship.

Ignore feedback about race conditions.

app.js
```diff
@@ -32,6 +32,7 @@ var ApplicationClients = require('./clients/');
 var ExitNode = require('./exit');
 var EntryNode = require('./entry');
 
+// our call SLA is 30 seconds currently
 var DRAIN_DEADLINE_TIMEOUT = 30 * 1000;
```
Raynos: just use `self.clients.serviceProxy.drainTimeout`

PR #89, opened 2015-10-26
Improve partial range

jcorbin:
- `#computePartialRange` now works completely in its "scratch space" object, which it returns for maximum debug transparency
- `#computePartialRange` took over the interval shenanigans that `#connecToServiceWorkers` used to have to
- connection logic is now inline in `#refreshServicePeerPartially` preparing for future work out of #66 
- explicit check and error log for pathological case

r @kriskowal @Raynos

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -567,6 +567,8 @@ function refreshServicePeerPartially(serviceName, hostPort) {
         return;
     }
 
+    var affineWorkers = self.getAffineWorkers(serviceName, range);
+
     self.logger.info('Refreshing service peer affinity', self.extendLogInfo({
         serviceName: serviceName,
         serviceHostPort: hostPort,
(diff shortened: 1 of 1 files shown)


Raynos:
legit.

service-proxy.js
```diff
@@ -494,60 +494,80 @@ ServiceDispatchHandler.prototype.computePartialRange =
...
+        start: NaN,
+        stop: NaN,
+        length: NaN,
+        affineWorkers: []
+    };
+
+    // Obtain and sort the affine worker and relay lists.
+    range.relays = Object.keys(self.egressNodes.exitsFor(serviceName));
+    range.relays.sort();
+    range.workers = serviceChannel.peers.keys();
```
jcorbin: hey so... isn't this keys array shared on the peers collection? and then we just mutable sort it? and log it? and .... :rage:

service-proxy.js
```diff
@@ -494,60 +494,80 @@ ServiceDispatchHandler.prototype.computePartialRange =
...
-        relayIndex: relayIndex,
-        relays: relays,
-        workers: workers,
-        length: length
-    };
+    range.stop = (range.start + range.length) % range.workers.length;
+
+    if (range.start === range.stop) {
+        // fully connected
+        range.affineWorkers = range.workers; // XXX .slice(0)?
```
kriskowal: .slice() would be superfluous. We will treat the array as immutable. Can remove comment.

service-proxy.js
```diff
@@ -564,34 +592,17 @@ function refreshServicePeerPartially(serviceName, hostPort) {
...
-        var peer = self.getServicePeer(serviceName, workers[i]);
+    for (var i = 0; i < range.affineWorkers.length; i++) {
+        peer = self._getServicePeer(chan, range.affineWorkers[i]);
         if (!peer.isConnected('out')) {
             peer.connectTo();
         }
     }
+
+    // TODO Drop peers that no longer have affinity for this service, such
+    // that they may be elligible for having their connections reaped.
```
kriskowal: This comment no longer applies, right?

kriskowal:
:ok:

PR #90, opened 2015-10-26
Ensure connected out and drain in

jcorbin:
Factor out ensurePeerConnected method, and drain the extra incoming connection in it.

r @rf @Raynos


service-proxy.js
```diff
@@ -494,6 +492,68 @@ function deletePeerIndex(serviceName, hostPort) {
...
+                'failed to ensure outgoing connection to service peer',
+                self.extendLogInfo({
+                    error: err,
+                    peerHostPort: peer.hostPort,
+                    refreshReason: reason
+                }));
+            return;
+        }
+        peer.drain({
+            reason: reason,
```
Raynos: We either need to communicate to every one to expect declined messages about draining.

Or make a change in the hyperbahn client in all languages to silence the draining error frame log.

Raynos:
lgtm. May cause hilarity / confusion.

service-proxy.js
jcorbin: Yup, that's a good educational moment to have anyhow, since we'll drain and decline during graceful shutdown anyhow, which we want them to handle.

Plus partial affinity in #66 will Soon ™ drain ex-affinity peers, so clients just really need to get used to this.

jcorbin:
Marking this as WIP until I get the tests passing, this is a nice to have, but not critical towards #66

PR #91, opened 2015-10-26
Affinity change vs refresh

jcorbin:
- add new `connectedServicePeers` index
- which lets us do a shorter fast path for refreshing an existing peer
- setting the stage to later to more work in the "something actually changed" path, such as disconnecting from ex-affine peers per the prior TODO

r @Raynos @kriskowal

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -472,9 +472,7 @@ function refreshServicePeer(serviceName, hostPort) {
     // The old way: fully connect every egress to all affine peers.
     self.addPeerIndex(serviceName, hostPort);
     var peer = self.getServicePeer(serviceName, hostPort);
-    if (!peer.isConnected('out')) {
-        peer.connectTo();
-    }
+    self.ensurePeerConnected(peer, 'service peer refresh');
 };
 
 ServiceDispatchHandler.prototype.addPeerIndex =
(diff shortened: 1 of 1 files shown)


jcorbin:
Disentangled from #90, since its in test-break hell.

service-proxy.js
```diff
@@ -465,35 +467,53 @@ function refreshServicePeer(serviceName, hostPort) {
...
 function deletePeerIndex(serviceName, hostPort) {
     var self = this;
 
+    deleteIndexEntry(self.connectedServicePeers, serviceName, hostPort);
     deleteIndexEntry(self.knownPeers, hostPort, serviceName);
 };
 
+ServiceDispatchHandler.prototype.ensurePeerConnected =
+function ensurePeerConnected(serviceName, peer, reason, now) {
+    var self = this;
```
kriskowal: TODO use reason

Raynos:
Reviewed things on commits.

kriskowal:
:ok:

service-proxy.js
jcorbin: indeed, it's here to be used by a future `peer.drain` such as #90

Raynos:
lgtm.

PR #93, opened 2015-10-27
Make hyperbahn compatible with batchStats

Raynos:
We made a breaking change in tchannel. I've reduced
the coupling to tchannel in hyperbahn.

Related tchannel PR: https://github.com/uber/tchannel-node/pull/139

r: @jcorbin @rf


Raynos:
Tests will fail without tchannel PR linked.

ShanniLi:
:+1:

PR #96, opened 2015-10-28
Factor IntervalScan out of ServiceDispatchHandler peer reap

jcorbin:
Factor IntervalScan out of ServiceDispatchHandler peer reaping

r @kriskowal @Raynos

### DIFF
--- a/lib/interval-scan.js
+++ b/lib/interval-scan.js
@@ -1,0 +1,214 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var assert = require('assert');
+var inherits = require('util').inherits;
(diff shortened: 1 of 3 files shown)


service-proxy.js
```diff
@@ -112,8 +113,25 @@ function ServiceDispatchHandler(options) {
...
+            return peersToReap;
+        },
+        self.channel.timers
+    );
+    self.peerReaper.runBeginEvent.on(function onPeerReapBegin(run) {
+        self.logger.info('reaping dead peers', self.extendLogInfo({
+            numPeersToReap: run.keys.length
+        }));
+    });
+    self.peerReaper.start(); // TODO: should IntervalScan be self-starting?
```
kriskowal: ServiceProxy itself starts on construction and stops on destruction. Perhaps, having start/stop methods should be a thing you build so you can pause/resume. destroy should stop regardless.

service-proxy.js
```diff
@@ -112,8 +113,25 @@ function ServiceDispatchHandler(options) {
...
+            self.knownPeers = Object.create(null);
+            return peersToReap;
+        },
+        self.channel.timers
+    );
+    self.peerReaper.runBeginEvent.on(function onPeerReapBegin(run) {
+        self.logger.info('reaping dead peers', self.extendLogInfo({
+            numPeersToReap: run.keys.length
+        }));
+    });
```
kriskowal: Usage looks good.

service-proxy.js
```diff
@@ -112,8 +110,53 @@ function ServiceDispatchHandler(options) {
...
+                    delete self.channel.subChannels[serviceName];
+                    self.rateLimiter.removeServiceCounter(serviceName);
+                    self.rateLimiter.removeKillSwitchCounter(serviceName);
+                }
+            }
+        },
+        // TODO: would be convenient to be able to pass an object or a function
+        // in cases like this
+        function getExitServices() {
+            return self.exitServices;
```
kriskowal: Consider capturing a snapshot.

jcorbin: Added a guard to IntervalScanRun#nextItem to skip undefined entries, I'd rather not introduce a copy into places like this where they didn't exist before; also I actively want to get a, maybe newer, lastRefresh by the time I get around to calling `maybePurgeEachService`

jcorbin:
I'm not 100% happy with the new constructor bloat, how about one or more .setupX methods?

jcorbin:
Punting the other two use cases to future PRs.

Also deferring any new Features like observability to future work

service-proxy.js
```diff
@@ -112,8 +112,26 @@ function ServiceDispatchHandler(options) {
     self.peersToReap = Object.create(null);
     self.knownPeers = Object.create(null);
 
-    self.reapPeersTimer = null;
-    self.reapPeersPeriod = options.reapPeersPeriod || DEFAULT_REAP_PEERS_PERIOD;
+    self.peerReaper = new IntervalScan(
+        'peer-reap',
```
Raynos: You are my friend.

lib/interval-scan.js
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    if (self.callback) {
+        self.callback();
+        self.callback = null;
+    }
+};
+
+IntervalScanRun.prototype.clearImmed =
+function clearImmed() {
```
kriskowal: typo: misspelling of Immediate

Raynos:
#isvalid

lib/interval-scan.js
Raynos: 80 characters.

Raynos: Immediate.

kriskowal:
:ok:

PR #97, opened 2015-10-29
test: improve test with more gaurds

Raynos:
This is needed because we introduced more errors
upstream and will avoid breaking master.

r: @jcorbin @kriskowal @rf


kriskowal:
ok

Raynos:
Test flap lol.

PR #98, opened 2015-10-29
Improve peer reap test

jcorbin:
I lost my mind trying to reason about this test and work with it towards #96, this is the preparatory result.

r @kriskowal @Raynos


kriskowal:
:ok:

PR #99, opened 2015-10-29
ServiceDispatchHandler#reapSinglePeer: improve skip check

jcorbin:
This change should be currently a noop, but this is closer to the source of truth and allows us to shift the generational rotation semantics.

r @kriskowal


kriskowal:
:ok:

PR #102, opened 2015-10-30
Enable the max tombstone ttl feature via flipr

Raynos:
r: @jcorbin @rf


jcorbin:
lgtm (rebased it for you)

PR #103, opened 2015-10-30
Create docs folder and index.md

Lupie:
Add initial doc template

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,0 +1,24 @@
+Hyperbahn
+=========
+
+Hyperbahn is a service discovery and routing network that allows simple, reliable
+communication with any service without any knowledge of where that service runs.
+
+Hyperbahn supports the TChannel RPC protocol and works with Ringpop's gossip protocol
+and consistent hash ring topology.
+
+Hyperbahn and TChannel clients are currently supported in Go, Python, and Node.js.
+
+* Go
+    * [Guide](go-guide.md)
+    * [API Documentation](http://godoc.org/github.com/uber/hyperbahn-go)
+
+
+* Node
+    * [Guide](http://hyperbahn-node.readthedocs.org/en/latest/GUIDE/)
+    * [API Documentation](http://hyperbahn-node.readthedocs.org/en/latest/)
+
+
+* Python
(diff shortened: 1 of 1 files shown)


blampe:
lgtm cc @Raynos @kriskowal @jcorbin @mranney

docs/index.md
```diff
@@ -0,0 +1,24 @@
...
+communication with any service without any knowledge of where that service runs.
+
+Hyperbahn supports the TChannel RPC protocol and works with Ringpop's gossip protocol
+and consistent hash ring topology.
+
+Hyperbahn and TChannel clients are currently supported in Go, Python, and Node.js.
+
+* Go
+    * [Guide](go-guide.md)
+    * [API Documentation](http://godoc.org/github.com/uber/hyperbahn-go)
```
prashantv: This link should probably be:
https://godoc.org/github.com/uber/tchannel-go/hyperbahn
or
https://godoc.org/github.com/uber/tchannel-go

docs/index.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+* Node
+    * [Guide](http://hyperbahn-node.readthedocs.org/en/latest/GUIDE/)
+    * [API Documentation](http://hyperbahn-node.readthedocs.org/en/latest/)
```
Raynos: do we want to point to "hyperbahn-node" docs or "tchannel-node" docs ?

I feel like we should point to tchannel documentation.

Lupie: These links don't work yet. You don't think we'd want docs specifically for
Hyperbahn, as its own product?

blampe: We don't have hyperbahn implementations in other languages, so it makes sense to just link to each language's tchannel implementation.

blampe: lgtm other than the links

Raynos: @Lupie Currently the documentation for the "hyperbahn client" that developers use lives in tchannel. tchannel ships with thrift & hyperbahn integration so any documentation for "how do i talk to hyperbahn" lives in tchannel.

Any documentation for "how do I run hyperbahn locally" lives here.

There's definitely a bunch of overlap here that can be improved, but not quite sure what the best seperation is.

PR #108, opened 2015-11-02
Add assert comments to reap test

jcorbin:
Pulled out of #104, since I wasn't able to get it past tests, and am unsure why adding these comments seems to make travis sad.

r @Raynos

### DIFF
--- a/test/lib/collapsed-assert.js
+++ b/test/lib/collapsed-assert.js
@@ -89,3 +89,10 @@ CollapsedAssert.prototype.report = function report(realAssert, message) {
         realAssert[method].apply(realAssert, command);
     }
 };
+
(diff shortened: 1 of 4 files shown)


test/lib/collapsed-assert.js
```diff
@@ -89,3 +89,10 @@ CollapsedAssert.prototype.report = function report(realAssert, message) {
         realAssert[method].apply(realAssert, command);
     }
 };
+
+CollapsedAssert.prototype.comment =
+function comment(msg) {
+    var self = this;
+
+    self._commands.push(['comment', msg]);
```
jcorbin: this part was written as dictation from @Raynos, and I have not independently verified it

Raynos:
lgtm.

PR #110, opened 2015-11-03
Partial affinity cleanups

jcorbin:
- add some more docs
- add a "neither connected, nor disconnected" mode to addPeerIndex
- skip logging or doing other work if there's no change under partial refresh
- drop an unused arg to `#changeToExit`

r @Raynos @rf

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -103,6 +103,11 @@ function ServiceDispatchHandler(options) {
      * peersToReap           :: Map<hostPort, lastRefresh>
      * knownPeers            :: Map<hostPort, lastRefresh>
      * connectedServicePeers :: Map<serviceName, Map<hostPort, lastRefresh>>
+     *
+     * On every advertise knownPeers is updated.
+     *
+     * However every reap period, knownPeers gets rolled over into peersToReap
+     * and emptied, so it represents the "peers seen this reap round"
      */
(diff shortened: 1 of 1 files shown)


service-proxy.js
```diff
@@ -494,9 +499,9 @@ ServiceDispatchHandler.prototype.addPeerIndex =
 function addPeerIndex(serviceName, hostPort, connected, now) {
     var self = this;
 
-    if (connected) {
+    if (connected === true) {
         addIndexEntry(self.connectedServicePeers, serviceName, hostPort, now);
-    } else {
+    } else if (connected === false) {
```
kriskowal: Please assert that connected is of type boolean.

service-proxy.js
```diff
@@ -644,6 +649,11 @@ function refreshServicePeerPartially(serviceName, hostPort, now) {
         }
     }
 
+    if (!toConnect.length) {
```
kriskowal: Please === 0

kriskowal:
:ok: pending :eyes: on ci

service-proxy.js
kriskowal: Offline: connected can be null. Please document what null means.

jcorbin: addressed by explicating null branch

PR #49, opened 2015-10-12
Add interface for caching sorted affine peers

kriskowal:
r @raynos @jcorbin

PR #111, opened 2015-11-03
Cache partial relays

jcorbin:
A simpler take on #49 towards #66, even if it is more monolith code for now.

r @Raynos @rf


jcorbin:
Depends on #110

Raynos:
seems legit.

PR #114, opened 2015-11-03
Factor out and generalize test cluster growth

jcorbin:
r @Raynos


jcorbin:
Tests pass, test only change, so merging.  Also verified in #115 that grow works the second time.

PR #116, opened 2015-11-03
Prune ex-service peers on membership change

jcorbin:
- add peer pruning facility
- add connectedPeerServices index so that we can correctly handle multi-tenant services
- implement a long standing TODO in exit -> forward transition under membership change

r @Raynos @rf

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -41,6 +41,7 @@ var DEFAULT_MIN_PEERS_PER_WORKER = 5;
 var DEFAULT_MIN_PEERS_PER_RELAY = 5;
 var DEFAULT_STATS_PERIOD = 30 * 1000; // every 30 seconds
 var DEFAULT_REAP_PEERS_PERIOD = 5 * 60 * 1000; // every 5 minutes
+var DEFAULT_PRUNE_PEERS_PERIOD = 2 * 60 * 1000; // every 2 minutes
 
 // our call SLA is 30 seconds currently
 var DEFAULT_DRAIN_TIMEOUT = 30 * 1000;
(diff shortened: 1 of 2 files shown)


service-proxy.js
```diff
@@ -51,6 +52,8 @@ var RATE_LIMIT_KILLSWITCH = 'killswitch';
 
 var CN_HEADER_BUFFER = new Buffer('cn');
 
+// TODO: audit all peer.drain for dupe
```
jcorbin: need TODO this before merging

service-proxy.js
```diff
@@ -506,8 +537,12 @@ function addPeerIndex(serviceName, hostPort, connected, now) {
 
     if (connected === true) {
         addIndexEntry(self.connectedServicePeers, serviceName, hostPort, now);
+        addIndexEntry(self.connectedPeerServices, hostPort, serviceName, now);
+        // TODO maybe delete self.peersToPrune[peer.hostPort];
```
jcorbin: reviewers, please pay attn to this part and provide feedback

service-proxy.js
```diff
@@ -979,6 +1047,80 @@ function setReapPeersPeriod(period) {
...
+    if (peer.draining) {
+        self.logger.info('skipping peer prune drain, already draining', self.extendLogInfo({
+            peer: peer.hostPort,
+            priorDrainReason: peer.drainReason
+        }));
+        return;
+    }
+
+    // TODO: stat?
+    self.logger.info('draining ex-affinity peer', self.extendLogInfo({
```
jcorbin: need to fix this log message, now that I've separated this feature out of ex-affinity

service-proxy.js
```diff
@@ -992,6 +1134,16 @@ function reapSinglePeer(hostPort, serviceNames) {
         return;
     }
 
+    if (peer.draining) {
+        // TODO: assuming that it's an ex-affinity drain, would be better to
```
jcorbin: my thinking on how to do this so far:
- we really need to reify the drain lifecycle of a peer (maybe also of channel)
- i.e. `peer.draining` should be an object owning the lifecycle, with properties such as
  - `peer.draining.reason` formerly peer.drainReason
  - `peer.draining.timeout` formerly peer.drainTimeout
  - `peer.draining.callback` formerly not available
- then we could branch on say the identity of `peer.draining.callback`
- and we could do such things an `peer.draining.cancel();` or...
- `peer.draining.update(/* new reason, direction, callback, etc, to change the course of a drain while it's still ongoing */);`

jcorbin:
This lays the groundwork for pruning ex-affinity peers in #66.

jcorbin:
Now depends on #130

jcorbin:
This is now passing tests, and ready for review after #133.

jcorbin:
Reviewed in person, merging

PR #117, opened 2015-11-04
Fix the peer reaper.

Raynos:
We had a bug in interval_scan; I added a test that the
peer reaper actually runs over real time.

r: @jcorbin @kriskowal @rf

### DIFF
--- a/lib/interval-scan.js
+++ b/lib/interval-scan.js
@@ -52,6 +52,7 @@ function IntervalScan(options) {
     self.getCollection = options.getCollection;
     self.timers = options.timers;
 
+    self.running = false;
     self.timer = null;
     self.theRun = null;
 
@@ -68,7 +69,7 @@ function setInterval(interval) {
     var self = this;
 
     if (self.interval === interval) {
-        if (!self.timer) {
+        if (self.running) {
             self.setTimer();
         }
         return;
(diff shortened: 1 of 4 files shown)


lib/interval-scan.js
```diff
@@ -68,6 +68,9 @@ function setInterval(interval) {
     var self = this;
 
     if (self.interval === interval) {
+        if (!self.timer) {
```
kriskowal: You can rely on setTimer to early return if !self.timer.

jcorbin: :+1: , just call self.setTImer()

test/register/peer-reaper-runs.js
```diff
@@ -0,0 +1,77 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var setTimeout = require('timers').setTimeout;
+
+var allocCluster = require('../lib/test-cluster.js');
+
+allocCluster.test('can register', {
```
jcorbin: should change this to be more descriptive

jcorbin:
lgtm now

kriskowal:
:ok:

jcorbin:
The regression was caused by #96 which generalized the pattern used by the peer reaper, which we then went on to use in #100, #101, and the upcoming #116.

PR #124, opened 2015-11-05
Fix detached publish

jcorbin:
This makes it easier to do a "detached" publish, like so:

```sh
$ git checkout v2.13.13
HEAD is now at 7695f10... 2.13.13

$ git cherry-pick -m1 master
[detached HEAD 2e7f285] Merge pull request #123 from uber/fix_bug
 Author: Joshua T Corbin <jcorbin@wunjo.org>
 Date: Thu Nov 5 10:59:30 2015 -0800
 1 file changed, 11 insertions(+), 1 deletion(-)

$ ./publish.sh patch
XXX this used to not work
```

Before this change, the `publish.sh patch` would blindly prepare and try to publish a package from the master branch, not the detached HEAD; now it will do so going forward.

r @Raynos @rf

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -30,7 +30,12 @@ fi
 
 npm version "$1"
 
-git push origin master --tags
+if head_ref=$(git symbolic-ref HEAD 2>/dev/null); then
+    branch_name=${head_ref##*/}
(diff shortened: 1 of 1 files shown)


rf:
cool :+1:

PR #127, opened 2015-11-05
remote-config property for peer choice with heap

rf:
r @Raynos @jcorbin

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -495,5 +495,7 @@ ApplicationClients.prototype.updateKillSwitches = function updateKillSwitches() 
 ApplicationClients.prototype.updatePeerHeapEnabled = function updatePeerHeapEnabled() {
     var self = this;
     var peerHeapConfig = self.remoteConfig.get('peer-heap.enabled.services', {});
-    self.serviceProxy.setPeerHeapEnabledServices(peerHeapConfig);
+    var peerHeapGlobalConfig = self.remoteConfig.get('peer-heap.enabled.global', false);
+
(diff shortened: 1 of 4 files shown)


clients/index.js
```diff
@@ -479,3 +480,16 @@ ApplicationClients.prototype.updateKillSwitches = function updateKillSwitches()
...
 };
+
+ApplicationClients.prototype.updatePeerHeapEnabled = function updatePeerHeapEnabled() {
+    var self = this;
+    var peerHeapConfig = self.remoteConfig.get('peer-heap.enabled.services', {});
+    var keys = Object.keys(peerHeapConfig);
+    for (var i = 0; i < keys.length; i++) {
+        var serviceName = keys[i];
+        var enabled = peerHeapConfig[serviceName];
+        if (self.channel.subChannels[serviceName]) {
```
Raynos: This is not correct. ServiceProxy will allocate sub channels dynamically.

Please call a method on ServiceProxy and do it properly there.

service-proxy.js
```diff
@@ -1250,6 +1255,14 @@ function extendLogInfo(info) {
     return info;
 };
 
+ServiceDispatchHandler.prototype.setPeerHeapEnabledServices =
+function setPeerHeapEnabledServices(peerHeapEnabledServices) {
+    var self = this;
+
+    assert(typeof peerHeapEnabledServices === 'object');
+    self.peerHeapEnabledServices = peerHeapEnabledServices;
```
Raynos: you still have to loop and mutate any existing sub channels.

Raynos:
#legit once tests pass.

PR #128, opened 2015-11-05
test: Remove flap by improving assert

Raynos:
This was flapping because non-deterministic log statement.
I added another case to catch the connection error
shape.

Observation: We could use an `extendLogInfo()` like
consistency for fields on error instance or stop asserting
on error fields and continue asserting on meta fields.

r: @jcorbin @rf


jcorbin:
lgtm... once tests pass

rf:
same :+1:

PR #129, opened 2015-11-05
Configure maximum relay ttl with flipr.

Raynos:
This allows us to drive it from flirp. I'm defaulting to
2 minutes for #extrafierce.

r: @jcorbin @rf


rf:
GREAT

PR #133, opened 2015-11-06
ServiceDispatchHandler: inline #addPeerIndex

jcorbin:
Now that we have the helper functions `addIndexEntry` and `deleteIndexEntry`,
it's not that much code anymore.

Further, inlining it leads to simpler code in #116, since otherwise the
temptation is to add yet another argument (4th) to `#addPeerIndex` to control
whether or not we should add/delete to/from the new secondary indices.

r @raynos @rf


jcorbin:
#fierce merging this since it's an isomorphic change, to unblock review of #116

PR #134, opened 2015-11-06
ServiceDispatchHandler: deafult the peer reaper on at 5min interval

jcorbin:
The peer reaper has proven itself, and accidentally not having it on
significant downside.  Since we've been running it in production at 5min
interval, let's make that the default.

r @raynos @rf


rf:
:+!:

PR #136, opened 2015-11-09
WIP: v0 of test-cluster child-process

Raynos:
Very rough work towards a test-cluster backed by a child-process.

Here are some wall clock numbers:
- test/health.js with embedded cluster: 1s
- test/health.js with child process cluster: 4-5s

I suspect the overhead is spawning a single child process per
test which hits the `require` and JIT warmup penalty every single
time.

The actual IPC communication and public interface of TestCluster
looks achievable with this approach but just requires a LOT more
code to be written to achieve feature parity.

Passing Tests:
- Only `test/health.js` has been verified.
- All other tests are unknown but will probably fail.

TODO:
- [ ] Performance: Use a long lived child process with an open IPC
     channel which we tell to bootstrap() and close() multiple
     test clusters to minimize child process overhead.
- [ ] Interface compatibility: Implemented all of the other features 
     in the public interface like dummies, namedRemotes, testRing, 
     cluster assertions, etc.


Raynos:
Also note; that to achieve child process caching. We will have to do some excessive coupling / integration with `tape` itself.

Ideally we lazily spawn a single child process and tear it down when the last test has finished. I think the best approximation would be to increment a global counter in `TestClusterProxy constructor` and decrement it in the end of `TestClusterProxy#close` after calling the callback. Every time the counter is zero; set a timer for 100ms and if the counter is still zero assume that the last test finished. 

Alternatively we can somehow `.unref()` the cached child process.

Raynos:
It's probably best to talk about this in person rather then "review" all of this code.

r: @jcorbin @rf @kriskowal

Raynos:
RIP

PR #140, opened 2015-11-11
Improve partial affinity refresh

jcorbin:
Changes made while debugging #66

- add audit against computed partial range
- call `#ensurePeerDisconnected` for symmetry and more consistency
- a bit more log info

r @Raynos @rf

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -698,6 +698,25 @@ function refreshServicePeerPartially(serviceName, hostPort, now) {
         // Mark known peers, so they are candidates for future reaping
         addIndexEntry(self.knownPeers, peer.hostPort, serviceName, now);
 
+        // TODO: this audit shouldn't be necessary once we understand and fix
+        // why it was needed in the first place
+        var range = self.getPartialRange(serviceName, 'refresh partial peer audit');
+        if (range) {
+            var shouldConnect = range.affineWorkers.indexOf(hostPort) >= 0;
+            var isConnected = !!connected;
+            if (isConnected !== shouldConnect) {
+                self.logger.warn('partial affinity audit fail', self.extendLogInfo({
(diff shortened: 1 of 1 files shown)


service-proxy.js
```diff
@@ -680,12 +698,34 @@ function refreshServicePeerPartially(serviceName, hostPort, now) {
...
+            if (isConnected !== shouldConnect) {
+                self.logger.warn('partial affinity audit fail', self.extendLogInfo({
+                    serviceName: serviceName,
+                    serviceHostPort: hostPort,
+                    isConnected: isConnected,
+                    shouldConnect: shouldConnect,
+                    connectedPeers: connectedPeers,
+                    partialRange: range
+                }));
+                connected = now;
```
jcorbin: @Raynos / @kriskowal this is the crucial change that fixes the heisen 1-3 fail we get out of `test/forward/dead-remote-reaped.js` under #66;  the failure is that some (between 1 and 3) of the resurrected pears end up having no outgoing connections made to them after resurrection.  Interesting note for other readers, we never see no failures without this; the test always provokes this condition, all that is indeterminate is how many resurrected peers are affected.

service-proxy.js
```diff
@@ -680,12 +698,34 @@ function refreshServicePeerPartially(serviceName, hostPort, now) {
...
+            }
+        }
+
         if (connected) {
             self.ensurePeerConnected(serviceName, peer, 'service peer affinity refresh', now);
+        } else {
+            self.ensurePeerDisconnected(serviceName, peer, 'service peer affinity refresh', now);
```
jcorbin: This didn't fix or harm anything, it just seems more correct to me.

kriskowal:
:ok:

PR #141, opened 2015-11-11
Improve some overly log-sensitive tests

jcorbin:
Towards #66:

- partial affinity info logs are now a normal part of cluster start up...
- ...so every single user of test cluster provokes them, even while bootstrapping
- so I chose to whitelist it in test cluster directly

I then found:
- the heap test was rather particular about seeing exactly its one log
- we had this forwarding test that also was sad about random new logs

Perhaps we should have the test-cluster go a step further and drop these logs
so that tests don't have to be modified to accommodate them?

r @raynos


Raynos:
Can we try https://github.com/Raynos/debug-logtron/pull/2 ?

PR #146, opened 2015-11-12
Use peer.drain to remove unadvertised peers

jcorbin:
- plus more info logging when we decide to not remove a multi-tenant peer

r @raynos


jcorbin:
Also @Raynos , this puts us one more prod usage and then all the test uses of `ready-signal` away from dropping another dep.

Raynos:
Story checks out. +1

PR #147, opened 2015-11-12
Cache partial ranges

jcorbin:
After deploying and canarying #66 we saw CPU saturation due to
`Array.prototype.sort`.  Caching the sorted workers should help since the
sorted relays are already cached.

cc @raynos @rf

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -100,7 +100,6 @@ function ServiceDispatchHandler(options) {
      * hostPort              :: string
      * lastRefresh           :: number // timestamp
      * partialRanges         :: Map<serviceName, PartialRange>
-     * relaysFor             :: Map<serviceName, List<hostPort>>
      * exitServices          :: Map<serviceName, lastRefresh>
      * peersToReap           :: Map<hostPort, lastRefresh>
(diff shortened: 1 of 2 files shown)


service-proxy.js
```diff
@@ -595,35 +604,33 @@ ServiceDispatchHandler.prototype.getPartialRange =
...
+    }
+
+    if (!partialRange.isValid()) {
         // This should only occur if an advertisement loses the race with a
         // relay ring membership change.
         self.logger.warn('Relay could not find itself in the affinity set for service', self.extendLogInfo({
             serviceName: serviceName,
             reason: reason,
-            partialRange: range
+            partialRange: partialRange
```
jcorbin: Dear reviewers:
- I could add the `.extendLogInfo` idiom to partialRange and use it here, thoughts?

rf: :+1:

rf:
lgtm, feel free to add `PartialRange#extendLogInfo`

Raynos:
Seems legit.

I hope you got all the cache invalidation hooks :p

PR #148, opened 2015-11-12
Downgrade erorr log about empty affine workers

jcorbin:
Turns out that this is a frequent edge case during transition times.  Full
explanation has not been nailed yet, but downgrading to a warn since this is
only a problem of it persists; what we saw was a brief blip of these at feature
toggle time.

r @raynos @rf


Raynos:
+1

rf:
:+1:

PR #150, opened 2015-11-13
WIP WIP: Add the ability to blacklist a service instance

Raynos:
This is a WIP PR to add the foundation for blacklisting
a service instance.
- Update the thrift file
- Create a new hyperbahn-service.js that implements blacklist;
  eventually we'll move all of handler.js into it and
  clean it up.
- Create a BlacklistInstance that manages the state and the
  timing aspects.
- Leave TODOs about HyperbahnServices interacts with
  EgressNodes & ServiceProxy.

TODO: 
- [ ] tests
- [ ] remove & drain peer in ServiceProxy
- [ ] fanout on EgressNodes change
- [ ] validation & other features

r: @jcorbin @rf

### DIFF
--- a/handler.js
+++ b/handler.js
@@ -32,6 +32,8 @@ var TChannelJSON = require('tchannel/as/json');
 var TChannelThrift = require('tchannel/as/thrift');
 var TChannelEndpointHandler = require('tchannel/endpoint-handler');
 
+var HyperbahnService = require('./hyperbahn-service.js');
+
 var MAX_RELAY_AD_ATTEMPTS = 2;
 var RELAY_AD_RETRY_TIME = 1 * 1000;
 var RELAY_AD_TIMEOUT = 500;
@@ -88,6 +90,11 @@ function HyperbahnHandler(options) {
         source: thriftSource
     });
 
+    self.hyperbahnService = new HyperbahnService({
+        channel: self.channel,
+        egressNodes: options.egressNodes
+    });
+
     // TODO replace JSON with real bufrw handlers for this
     self.tchannelJSON.register(self, 'ad', self,
         self.handleAdvertise);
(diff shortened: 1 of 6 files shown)


hyperbahn-service.js
```diff
@@ -0,0 +1,71 @@
...
+    // TODO: listen to egressNodes and fanout to re-blacklist ?
+}
+
+HyperbahnService.prototype.blacklist =
+function blacklist(req, headers, body, cb) {
+    var self = this;
+
+    // TODO: validate body.serviceName
+    // TODO: validate body.instanceHostPort
+    // TODO: validate body.timeToBan
```
jcorbin: it's thrift, how are they not already validated

jcorbin:
Please "ban" something, e.g. "banPeer", "banWorker", etc

jcorbin:
to me "blacklist" is indefinite without expiry, where-as "ban" seems much more sensible to be time-bounded.

hyperbahn-service.js
Raynos: edge cases:

 - empty string
 - non-valid string format host:port
 - negative number / zero.

handler.js
```diff
@@ -127,11 +142,28 @@ HyperbahnHandler.prototype.type = 'hyperbahn.advertisement-handler';
 */
 HyperbahnHandler.prototype.handleAdvertise =
 function handleAdvertise(handler, req, arg2, arg3, cb) {
-    handler.sendRelays(req, arg2, arg3, 'relay-ad', cb);
+    var services = arg3.services;
+    var fanoutServices = [];
+
+    for (var i = 0; i < services.length; i++) {
```
rf: Could we do this logic in `#sendRelays` instead? There's already a loop over the host ports there that adds them to a new data structure, would be nice to only do that once.

Raynos:
RIP

PR #152, opened 2015-11-13
Add missing parital affinty call to unadvertise

jcorbin:
We previously did nothing to partial affinity on unadvertise.

r @raynos @rf


service-proxy.js
```diff
@@ -824,6 +825,19 @@ function removeServicePeer(serviceName, hostPort) {
     }
     serviceChannel.peers.delete(hostPort);
 
+    if (self.partialAffinityEnabled) {
+        var result = self.ensurePartialConnections(
```
jcorbin: this is all an analog of logic in `#refreshServicePeerPartially` circa L651

Raynos:
legit.

PR #154, opened 2015-11-13
A couple of trivial cleanups in ServiceDispatchHandler

jcorbin:
- consistent formatting on log sites for minimal diff churn going forward
- consistent naming of PartialRange object instances

cc @rf @raynos


Raynos:
legit.

PR #158, opened 2015-11-17
Test and fix PartialRange

jcorbin:
Less flaps, more correct.

r @rf @raynos @kriskowal


kriskowal:
Looking reasonable so far.

partial_range.js
```diff
@@ -93,11 +93,18 @@ function compute(relays, workers, now) {
...
-    this.length        = Math.ceil(this.minPeersPerWorker * this.ratio);  // how many peers we are going to connect to
-    this.length        = Math.max(this.minPeersPerRelay, this.length);    // please always have this many
-    this.length        = Math.min(this.workers.length, this.length);      // you can't have more than there are
-    this.start         = Math.floor(this.relayIndex * this.ratio);
-    this.stop          = Math.ceil(this.relayIndex * this.ratio + this.length) % this.workers.length;
+    this.length = Math.ceil(this.minPeersPerWorker * this.ratio);  // how many peers we are going to connect to
+    this.length = Math.max(this.minPeersPerRelay, this.length);    // please always have this many
+    this.length = Math.min(this.workers.length, this.length);      // you can't have more than there are
+
+    if (this.length >= this.workers.length) {
```
Raynos: Wait; if there are more hyperbahn workers then service workers we fully connect ?

What if k=120 and service=100 ?

Are we assuming that for the case that hyperbahn workers > service workers k<10 ?

This tradeoff seems reasonable to me; we should just be aware of it.

Raynos:
lgtttm.

PR #159, opened 2015-11-17
Add test cluster port symbolication

jcorbin:
cc @raynos @rf


rf:
super rad and neat :+1:

PR #160, opened 2015-11-17
Add a second audit path to partial affinity

jcorbin:
This was the source of (flap and then constant under #158) failure out of
`test/forward/dead-remote-reaped.js`.  As yet unexplained why either of our our
partial connected audit sites are necessary.

r @raynos @rf @kriskowal


Raynos:
legit; please remove the "LOL"

PR #166, opened 2015-11-20
Update hostports tests for thriftrw 3.0.1

malandrew:
PR to get tests for tchannel-node to pass after upgrading to thriftrw 3.0.1

### DIFF
--- a/endpoints/index.js
+++ b/endpoints/index.js
@@ -74,7 +74,11 @@ function setupEndpoints(clients, services) {
     /*eslint no-unused-vars: 0*/
     // thrift health endpoint
     var tchannelAsThrift = new TChannelAsThrift({
-        source: '// lul',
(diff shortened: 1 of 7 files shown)


kriskowal:
I I don’t think it will be adequate to fix this in hyperbahn. To appease the tests (and satisfy backward compatibility) we will need to address the issue in TChannel node by coercing the old usage.

malandrew:
so support `{source: ... }`? Can we just major version bump?

kriskowal:
@malandrew It’s become clear that major version bumps are expensive since we have to backport fixes to prior major versions of tchannel.

Raynos:
did this get merged somewhere else ?

Raynos:
I dont think this is necessary; going to close.

PR #167, opened 2015-11-20
add a socket inspector

Raynos:
We get really annoying errors in production because we
forgot to add an "error" listener to a Socket somewhere
in the process.

```
Error: read ECONNRESET
    at errnoException (net.js:904:11)
    at TCP.onread (net.js:558:19)
```

To be able to debug this I'm monkey patching the 
Socket.prototype.emit method to allow us to intercept it.

When we intercept it we do two things:

 - create a new wrapped error to get more stack frames
 - log everything about the socket that errored

r: @shannili @rf @jcorbin


ShanniLi:
shipit!

PR #168, opened 2015-11-20
Upgrade uncaught to latest

Raynos:
`uncaught-exception` now contains a new feature of printing
the current stack whenever an uncaught exception happens.

This helps debugging errors with really small stack traces
like any TCP exceptions.

r: @jcorbin @rf @shannili


ShanniLi:
:+1:

PR #170, opened 2015-11-24
Add a benchmark mode which spams peer refresh (the meat of advertise)

jcorbin:
Going forward it'd be great to have this be more "realistic" by actually going through the RPC handler path, but this was achievable and useful for tuning partial affinity.

r @Raynos @rf


Raynos:
:+1:

PR #172, opened 2015-11-26
Pare down and limit use of secondary indices

jcorbin:
- only update them under the partial affinity path
- drop one of them entirely in lieu of some smaller sub-objects on `PartialRange`
- fix a potential bug under `reapSinglePeer`

r @raynos @rf
cc @kriskowal


Raynos:
legit.

PR #173, opened 2015-11-27
rewrite traffic to global stats client

Raynos:
The cardinality of our metrics is insane. This change
removes the host/worker cardinality from our inbound
and outbound stats.

r: @rf @shannili @robskillington

### DIFF
--- a/clients/batch-stats.js
+++ b/clients/batch-stats.js
@@ -1,0 +1,116 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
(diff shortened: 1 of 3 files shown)


ShanniLi:
lgtm

clients/batch-stats.js
```diff
@@ -0,0 +1,116 @@
...
+        stat.name === 'tchannel.outbound.calls.per-attempt-latency' ||
+        stat.name === 'tchannel.outbound.calls.per-attempt.operational-errors' ||
+        stat.name === 'tchannel.outbound.calls.system-errors' ||
+        stat.name === 'tchannel.outbound.response.size' ||
+        stat.name === 'tchannel.outbound.calls.per-attempt-latency' ||
+        stat.name === 'tchannel.outbound.calls.success' ||
+        stat.name === 'tchannel.outbound.calls.per-attempt.app-errors' ||
+        stat.name === 'tchannel.outbound.response.size'
+    ) {
+        self.writeToStatsd(self.statsd.globalClient, stat);
```
robskillington: For all these string checks, wouldn't it be much faster to check if the stat is in a precomputed hash?  Would greatly reduce string comparisons.

Raynos: dictionary lookups are O(1) but the constant value but they are not free. I suspect that this check is cheaper.

PR #174, opened 2015-11-27
do not emit edge-rps stats at all

Raynos:
This removes a bunch of unique metrics that are hard to scale

r: @rf @jcorbin


jcorbin:
:+1:

PR #175, opened 2015-11-28
WIP: WIP The great process seperation

Raynos:
This is a double WIP. I'm toying around with how to actually
seperate these beasts.

I've tried to keep the tests passing on every commit and avoided
monster commits where possible.

The general approach is:
- split app.js into discovery/discovery-worker.js and
  router/routing-worker.js
- move tchannel, ringpop, serviceProxy out of clients/index.js
  and into the correct worker file
- introduce discovery/routing-bridge.js and router/discovery-bridge.js
  to provide an interface that will be implemented with IPC
- use tchannel/thrift for the actual IPC because tchannel/thrift is
  as fast as node-redis.

TODO:
- [ ] break service-proxy.js into two parts...
- [ ] add support for "hs" host header in relay so that discovery
  can route ringpop RPCs through the routing listening socket
- [ ] use IPC for the bridge but have both halves in one process
- [ ] have discovery spawn routing as child process
- [ ] Refactor tests... to support multi-process...
- [ ] #fierce deploy to canary if tests green.

cc @jcorbin @rf


Raynos:
More progress has been added to this PR.

The main unit of progress is that i've started breaking service-proxy.js into three independant parts:

 - `router/service-proxy-handler`. this implements `handleLazily()` and actually does all the work
 - `router/service-routing-table`. This exists inside the router process and will handle creating sub channels and configuring the peers & connections.
 - `service-proxy`. This exists in the discovery process and now only handles affinity, partial affinity, ringpop integration, advertise, refreshServicePeer, peer reaper, etc.

There are a couple of "shortcuts" that I've taken in service-routing-table by unsafely reaching directly into the discovery entity and the service-proxy still completely unsafely reaches into tchannel itself.

The service-proxy-handler is a good shape though with blocking, routing & rate limiting working. I've had to comment out the circuit breaker tests because relay-network simply needs to die, the circuit breaker tests have to be rewritten against test-cluster directly

The next unit of work is getting service-routing-table and service-proxy to correctly use the bridge interfaces to not cheat. I.e. define some message passing between the two rather then mutating the shared tchannel data structures.

Raynos:
RIP

PR #176, opened 2015-12-01
we should probably call join()

Raynos:


kriskowal:
Yes. We probably ought to call join(). :ok:

PR #177, opened 2015-12-01
service-proxy: implement routing delegate header

rf:
r @Raynos @jcorbin 

If there's a routing delegate we defer to it as soon as possible.

Should this actually hit the rate limiting logic, at least for total RPS?

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -432,7 +432,7 @@ function handleRequest(req, buildRes) {
                 }))
             );
             if (rd) {
-                buildRes().sendError('Busy', 'Routing delegate ' + rd + ' is rate-limited by the rps of ' + serviceLimit);
+                buildRes().sendError('Busy', 'Routing delegate ' + rd + ' is rate-limited by the service rps of ' + serviceLimit);
             } else {
                 buildRes().sendError('Busy', req.serviceName + ' is rate-limited by the service rps of ' + serviceLimit);
             }
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -302,11 +302,6 @@ function handleLazily(conn, reqFrame) {
 
     var rdBuf = res.value && res.value.getValue(RD_HEADER_BUFFER);
     var rd = rdBuf && rdBuf.toString();
-    if (rd) {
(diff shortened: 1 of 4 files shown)


Raynos:
Please also `handleRequest` for eager support.

service-proxy.js
```diff
@@ -350,9 +354,10 @@ function handleLazily(conn, reqFrame) {
         }
     }
 
-    var serviceChannel = self.channel.subChannels[serviceName];
+    // use the rd (routing delegate) or the serviceName if there was no rd set
+    var serviceChannel = self.channel.subChannels[rd || serviceName];
```
Raynos: The weird thing is that we double count the rate limiter. we shouldnt increment rate limit for service if we do not route to service :(

Raynos:
Weird that tests didnt fail without eager.

I think we default into lazy now and need to update the travis to re-run the entire test suite through eager.

Can you update the travis test config thingy ?

.travis.yml
```diff
@@ -6,7 +6,7 @@ before_install:
 env:
 # Get some of them configs
 - HYPERBAHN_REMOTE_TEST_CONFIG=
-- HYPERBAHN_REMOTE_TEST_CONFIG=test/config/with_lazy_relay.json
+- HYPERBAHN_REMOTE_TEST_CONFIG=test/config/without_lazy_relay.json
```
jcorbin: good catch, we did change the default recently, please break this out into a separate PR

service-proxy.js
```diff
@@ -376,13 +385,15 @@ function handleRequest(req, buildRes) {
         return;
     }
 
-    if (self.isBlocked(req.headers && req.headers.cn, req.serviceName)) {
+    var rd = req.headers && req.headers.rd;
```
jcorbin: can we call this actually `routingDelegate` and then add a `var nextService = routingDelegate || req.serviceName;` and use it throughout?

rf:
Split travis config fix into https://github.com/uber/hyperbahn/pull/180

Raynos:
Good ship it.

Raynos:
lgtm.

PR #178, opened 2015-12-01
Cleanup and fix benchmarks makefile

jcorbin:
- deduped all the argument classes
- lowered the pipeline values so that we don't provoke
  tchannel.socket.write-full errors
- pattern-ized the flame generation rules

r @raynos @rf


rf:
shur :+1:

PR #180, opened 2015-12-01
Fix travis config to run eager tests again

rf:
Also removes unneeded lazy-specific rate limiter test

r @jcorbin @Raynos


Raynos:
lgtm :)

jcorbin:
:ok: I'd prefer to keep config snippets on hand for both pro and con, but not a big deal

PR #181, opened 2015-12-02
Rework benchmarking

jcorbin:
- use JSON config fragments
- use a patterned Makefile


jcorbin:
Depends on https://github.com/uber/tchannel-node/pull/220 and https://github.com/uber/tchannel-node/pull/221 .

Raynos:
This is cool. #doit

PR #183, opened 2015-12-03
Improve the test remote config fixture

jcorbin:
- clearer tmp filename
- prettier JSON
- support passing thru a key-val array

r @raynos


Raynos:
lgtm.

PR #185, opened 2015-12-03
Restore caching of exit relays

jcorbin:
This gives ~150% increase in performance of partial affinity computation
under the `relay-ad-partial` benchmark config; this PR takes the bench rates
from ~2500/s to ~6200/s.

r @raynos @rf


jcorbin:
In fact, this makes the rate/s relay-ad-partial ~10% faster than relay-ad (the non-partial affinity ad-spammer).

Raynos:
Do we has flames ?

Raynos:
lgtm.

PR #186, opened 2015-12-03
Do less work to add/remove workers to/from PartialRange

jcorbin:
r @Raynos @rf

### DIFF
--- a/partial_range.js
+++ b/partial_range.js
@@ -81,6 +81,28 @@ function compute(relays, workers, now) {
     this.recompute(now);
 };
 
+PartialRange.prototype.addWorker =
+function addWorker(hostPort, now) {
+    var i = sortedIndexOf(this.workers, hostPort);
+    if (i >= 0) {
+        return;
+    }
+
+    this.workers.splice(~i, 0, hostPort);
+    this.recompute(now);
+};
+
+PartialRange.prototype.removeWorker =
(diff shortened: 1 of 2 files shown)


jcorbin:
This gets the torch difference between full and partial refresh peers to within a fraction of a percent.

service-proxy.js
```diff
@@ -1312,7 +1308,7 @@ function pruneSinglePeer(hostPort, pruneInfo) {
 };
 
 ServiceDispatchHandler.prototype.reapSinglePeer =
-function reapSinglePeer(hostPort, serviceNames) {
+function reapSinglePeer(hostPort, serviceNames, now) {
```
rf: `now` will be undefined, no? is it needed?

jcorbin: good catch, fixed

partial_range.js
```diff
@@ -81,6 +81,28 @@ function compute(relays, workers, now) {
...
 };
 
+PartialRange.prototype.addWorker =
+function addWorker(hostPort, now) {
+    var i = sortedIndexOf(this.workers, hostPort);
+    if (i >= 0) {
+        return;
+    }
+
+    this.workers.splice(~i, 0, hostPort);
```
Raynos: y u do dis.

Raynos:
Seems legit.

PR #188, opened 2015-12-04
Improve hyperbahn torching experience with better defaults

Raynos:
- Added more multiplicity for more data
 - Added noEndpointOverhead for more throughput
 - Added 15s delay instead of 10s for warmer process.

r: @jcorbin

### DIFF
--- a/benchmarks/config/torch.json
+++ b/benchmarks/config/torch.json
@@ -1,5 +1,6 @@
 {
     "torch": "relay",
     "torchTime": 10,
+    "multiplicity": 3,
     "noEndpointOverhead": true
 }
--- a/benchmarks/config/torch.json
+++ b/benchmarks/config/torch.json
@@ -1,6 +1,7 @@
 {
     "torch": "relay",
     "torchTime": 10,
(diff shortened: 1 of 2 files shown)


benchmarks/index.js
```diff
@@ -43,6 +43,7 @@ function HyperbahnBenchmarkRunner(opts) {
     }
 
     var self = this;
+    opts.torchDelay = opts.torchDelay || 15 * 1000;
```
jcorbin: what is this one for?

benchmarks/config/torch.json
```diff
@@ -1,4 +1,6 @@
 {
     "torch": "relay",
-    "torchTime": 10
+    "torchTime": 10,
+    "multiplicity": 3,
```
jcorbin: derp, good catch, this got missed out of #181

benchmarks/index.js
jcorbin: anyhow, if we keep it please:
- make the units be seconds in like `torchTime`
- add the default to `torch.json` for affordance

PR #191, opened 2015-12-04
Stop logging full connectedPeers arrays under audit failures

jcorbin:


Raynos:
This definitely works. lgtm.

PR #192, opened 2015-12-05
expose initHeaders on exit_connections_v1

Raynos:
This allows us to debug what connections we have open
and who is on the other side

r: @rf @shannili @kriskowal @jcorbin


rf:
fix lint

rf:
:+1: once travis

PR #197, opened 2015-12-09
Eliminate ability for ringpop and partial affinity to thrash

jcorbin:
r @raynos @rf


kriskowal:
:ok:

Raynos:
story checks out.

PR #199, opened 2015-12-10
Stop logging audit fail warns for the causal worker

jcorbin:
After feature flag transition, `path: "ensurePartialConnections: advertise"` is
the only class of continuing audit fails (the `path: "freshenPartialPeer"` ones
are an impulse at transition time).

Making this change so we can see how many are due to the worker that just
advertised, or not.

r @raynos @kriskowal


kriskowal:
:ok:

PR #202, opened 2015-12-10
Defer opening all peer connections.

jcorbin:
Connections are now opened in smeared batches every `100ms` (configurable by `peerConnecter.period`).

This will mitigate CPU load under heavy advertise spam, worker churn, or internal hyperbahn topology change.

r @Raynos @rf

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -460,6 +460,7 @@ function onRemoteConfigUpdate(changedKeys, forceUpdate) {
     self.updateReservoir(hasChanged, forceUpdate);
     self.updateReapPeersPeriod(hasChanged, forceUpdate);
     self.updatePrunePeersPeriod(hasChanged, forceUpdate);
+    self.updateConnectPeersPeriod(hasChanged, forceUpdate);
     self.updatePartialAffinityEnabled(hasChanged, forceUpdate);
     self.setMaximumRelayTTL(hasChanged, forceUpdate);
     self.updatePeerHeapEnabled(hasChanged, forceUpdate);
@@ -575,6 +576,15 @@ function updatePrunePeersPeriod(hasChanged, forceUpdate) {
     if (forceUpdate || hasChanged['peerPruner.period']) {
         var period = self.remoteConfig.get('peerPruner.period', 0);
         self.serviceProxy.setPrunePeersPeriod(period);
+    }
+};
+
+ApplicationClients.prototype.updateConnectPeersPeriod =
+function updateConnectPeersPeriod(hasChanged, forceUpdate) {
+    var self = this;
+    if (forceUpdate || hasChanged['peerConnecter.period']) {
+        var period = self.remoteConfig.get('peerConnecter.period', 100);
+        self.serviceProxy.setConnectPeersPeriod(period);
     }
 };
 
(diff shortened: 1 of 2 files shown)


clients/index.js
```diff
@@ -483,6 +484,13 @@ function updatePrunePeersPeriod() {
     self.serviceProxy.setPrunePeersPeriod(period);
 };
 
+ApplicationClients.prototype.updateConnectPeersPeriod =
+function updateConnectPeersPeriod() {
+    var self = this;
+    var period = self.remoteConfig.get('peerConnecter.period', 0);
```
Raynos: we shouldnt default to off.

If we cant talk to flipr we should default to like 100ms or something.

Raynos:
lgtm.

service-proxy.js
```diff
@@ -166,6 +168,34 @@ function ServiceDispatchHandler(options) {
...
+        getCollection: function getPeersToConnect() {
+            var peersToConnect = self.peersToConnect;
+            if (!Object.keys(peersToConnect).length) {
+                return null;
+            }
+            self.peersToConnect = Object.create(null);
+            return peersToConnect;
+        }
+    });
+    self.peerConnecter.runBeginEvent.on(function onPeerReapBegin(run) {
```
rf: `onPeerReapBegin`? why not `onPeerConnectorBegin`?

rf:
looks ok besides nitpick

Raynos:
I think this is under test framework hell.

service-proxy.js
```diff
@@ -918,6 +961,7 @@ function ensurePeerDisconnected(serviceName, peer, reason, now) {
 
     var peerServices = self.connectedPeerServices[peer.hostPort];
     if (!peerServices || isObjectEmpty(peerServices)) {
+        delete self.peersToConnect[peer.hostPort];
```
Raynos: is this sufficient ? The peer connector could be in it's "100ms setImmediate loop" right now. i.e. `peersToConnect` was already copied and its doing the smeared loop.

Raynos:
This looks legit.

service-proxy.js
jcorbin: Good question, I just updated the connectSinglePeer implementation to:
- stop if there's a prune note
- not create the peer if it doesn't exist
- stop if the peer, if it still exists, is draining

thoughts?

service-proxy.js
```diff
@@ -675,6 +705,36 @@ function ensurePeerConnected(serviceName, peer, reason, now) {
...
+function connectSinglePeer(hostPort, connectInfo) {
+    var self = this;
+
+    if (self.peersToPrune[hostPort]) {
+        return;
+    }
+
+    var serviceName = connectInfo.serviceName;
+    var serviceChannel = self.getServiceChannel(serviceName);
+    if (!serviceChannel) {
```
Raynos: I think this is broken... unless we allocate the peer synchronously.

Raynos: I see we do allocate the peer; so if the peer was de-allocated then we do not connect.

Are there any race conditions here where we remove peers. I dont think there are.

service-proxy.js
```diff
@@ -675,6 +705,36 @@ function ensurePeerConnected(serviceName, peer, reason, now) {
...
+        return;
+    }
+
+    var peer = serviceChannel.peers.get(hostPort);
+    if (!peer) {
+        return;
+    }
+
+    if (!peer.draining) {
```
Raynos: this seems broken; `if (peer.draining) return;` no negation.

Raynos:
This seems like a significant improvement :)

PR #205, opened 2015-12-11
ServiceDispatchHandler#ensurePartialConnections: defer all work

jcorbin:


rf:
:+1:

Raynos:
I think there is a lot of test code that assumes when ad() finishes then all the connections are open. 

We could just only denounce in ring change. I.e. Edge triggered relay ads immediately connect()

Ring triggered changes debounce.

PR #207, opened 2015-12-23
Bail out of egress Hyperbahn::discover calls for non-affine services

kriskowal:
r @Raynos @jcorbin @thanodnl

### DIFF
--- a/handler.js
+++ b/handler.js
@@ -397,7 +397,7 @@ function discover(handler, req, head, body, cb) {
 
     var svcchan = null;
     var myHost = handler.channel.hostPort;
-    if (exitHosts.indexOf(myHost) === -1) {
+    if (exitHosts.indexOf(myHost) < 0 && req.headers.cn !== 'hyperbahn') {
         // Since Hyperbahn is fully connected to service hosts,
         // any exit node suffices.
         svcchan = handler.channel.topChannel.handler.getOrCreateServiceChannel(serviceName);


kriskowal:
This is a first stab at this issue. I’m using the "cn" header in this PR, but if we elect to "fake" the CN with the original service to create a per-origin circuit breaker on the Hyperbahn channel, this will not work.

kriskowal:
We could alternately add a property to the IDL to distinguish a forwarded request. Not sure how Ringpop’s handle-or-forward resolved the issue cc @thanodnl

thanodnl:
@kriskowal HandleOrForward only 'solves' it when you do not overwrite the endpoint your self. HandleOrForwards defaults to forward to the `/proxy/req` endpoint on the `ringpop` channel. And the handler there does not forward anything. It is a really hard api to use correctly.

How we lately solved this in `ringpop-go` is by setting a [`ringpop-forwarded`](https://github.com/uber/ringpop-go/blob/master/forward/forwarder.go#L189-L197) in the application level header. The call is then sent to ring member that owns the key. Here a check is done to see if it is a forwarded call and if so it always executes it locally.

Lately I have been thinking that the node it is forwarded to should also check it believes it is the owner of the shardkey, if not send an error response back. But that is something that we are not really sure about how to handle at the moment. And it might be application specific what needs to be done. Some applications do not work if they execute things on the wrong machines, some applications have degraded performance so it will probably be a configuration option for application owners to specify.

handler.js
```diff
@@ -397,7 +397,7 @@ function discover(handler, req, head, body, cb) {
 
     var svcchan = null;
     var myHost = handler.channel.hostPort;
-    if (exitHosts.indexOf(myHost) === -1) {
+    if (exitHosts.indexOf(myHost) < 0 && req.headers.cn !== 'hyperbahn') {
```
thanodnl: As I understand it correctly the discover call will be executed locally when it has been called by hyperbahn ie. forwarded. What would the response be if it has been forwarded to the wrong machine because of ring rebalance that has not completely propagated? Will it just work, will you get an empty list back, or is there an error?

As I described in the chat about how ringpop solves it you might want to double check you are an exithost even if the call has been forwarded and throw an error if you are not an exitnode.

Raynos:
I think the correct solution is two methods

`discover()` and `discoverAffinity()`

Raynos:
Let's get this in and let's get a hotpatch build ready just in case.

Raynos:
Merging.

kriskowal:
Sounds good. We can make a better fix in a follow-up.

PR #208, opened 2015-12-23
Add a staging mode.

Raynos:
This adds a new entry point for tmux as well
as a Makefile.

r: @jcorbin @kriskowal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PUBLIC_IP=$(shell ifconfig eth0 | grep 'inet addr:' | cut -d: -f2 | awk '{ print $$1}')
+PUBLIC_IP=$(shell node scripts/public-ip.js)
 
 # Helper for running autobahn servers locally.
 run-local-%:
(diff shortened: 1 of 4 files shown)


hyperbahn-staging.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+set -e
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
jcorbin: `$(dirname $(readlink -f $0))` but only on GNU system... maybe what you have is The Way 
™ for Portability ™.

Makefile
```diff
@@ -1,4 +1,8 @@
+PUBLIC_IP=$(shell ifconfig eth0 | grep 'inet addr:' | cut -d: -f2 | awk '{ print $$1}')
```
jcorbin: - hardcoding eth0 isn't portable
- I bet "inet addr:" is specific to linux ifconfig

jcorbin:
lgtm modulo hacked shell scripting

PR #209, opened 2016-01-05
discovery: Fix Hyperbahn::discover to not have a loop

Raynos:
This fixes two issues:

 - Instead of doing an RPC loop have two methods, one
	which terminates
 - To avoid retry storms disable retries.

This PR also fixed a bunch of annoying testing things
in test-cluster at the top of the commit history, I can
break those out into a seperate PR.

r: @jcorbin @kriskowal

### DIFF
--- a/handler.js
+++ b/handler.js
@@ -447,10 +447,13 @@ function _forwardToRemoteDiscover(parent, svcchan, body, cb) {
 
     function handleForward(err, resp) {
         if (err) {
-            self.channel.logger.error('Failed to call discover API on exit node', {
-                error: err,
-                serviceName: serviceName
-            });
+            self.channel.logger.error(
+                'Failed to call discover API on exit node',
+                parent.extendLogInfo({
+                    error: err,
+                    serviceName: serviceName
+                })
+            );
             return cb(err, null);
         }
 
--- a/handler.js
+++ b/handler.js
@@ -438,8 +438,10 @@ function _forwardToRemoteDiscover(parent, svcchan, body, cb) {
             cn: 'hyperbahn'
         },
         parent: parent,
-        timeout: 5000,
-        timeoutPerAttempt: 500,
+        retryFlags: {
+            never: true
+        },
+        timeout: 1000,
         trace: false
     }), endpoint, null, body, handleForward);
 
--- a/handler.js
+++ b/handler.js
@@ -392,54 +392,64 @@ function discover(handler, req, head, body, cb) {
         return;
     }
 
-    var exitNodes = handler.egressNodes.exitsFor(serviceName);
-    var exitHosts = Object.keys(exitNodes);
+    var topChannel = handler.channel.topChannel;
+    var svcchan = topChannel.handler.getOrCreateServiceChannel(serviceName);
 
-    var svcchan = null;
-    var myHost = handler.channel.hostPort;
-    if (exitHosts.indexOf(myHost) < 0 && req.headers.cn !== 'hyperbahn') {
-        // Since Hyperbahn is fully connected to service hosts,
-        // any exit node suffices.
-        svcchan = handler.channel.topChannel.handler.getOrCreateServiceChannel(serviceName);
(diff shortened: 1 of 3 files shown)


handler.js
```diff
@@ -392,56 +402,96 @@ function discover(handler, req, head, body, cb) {
         return;
     }
 
-    var exitNodes = handler.egressNodes.exitsFor(serviceName);
```
kriskowal: nice

kriskowal:
lgtm

kriskowal:
can you explain why you’ve made the routing loop a remote configurable option?

handler.js
```diff
@@ -392,56 +402,96 @@ function discover(handler, req, head, body, cb) {
...
+};
+
+HyperbahnHandler.prototype._forwardToRemoteDiscover =
+function _forwardToRemoteDiscover(parent, svcchan, body, cb) {
+    var self = this;
+    var serviceName = svcchan.serviceName;
+
+    var endpoint = self.discoveryRouteMethod;
+
+    // Since Hyperbahn is fully connected to service hosts,
```
jcorbin: it's not about connected fully vs partially; it's because we fully relay all advertisements

handler.js
```diff
@@ -392,56 +402,96 @@ function discover(handler, req, head, body, cb) {
...
+        retryFlags: {
+            never: true
+        },
+        timeout: 1000,
+        trace: false
+    }), endpoint, null, body, handleForward);
+
+    function handleForward(err, resp) {
+        if (err) {
+            self.channel.logger.error('Failed to call discover API on exit node', {
```
jcorbin: warn imo

jcorbin: something.extendLogInfo(...)

package.json
```diff
@@ -32,7 +32,7 @@
     "my-local-ip": "1.0.0",
     "null-logtron": "2.2.0",
     "process": "0.10.0",
-    "process-reporter": "2.0.0",
+    "process-reporter": "2.1.0",
```
jcorbin: neat in passing upgrade; separate PR please

test/lib/test-cluster.js
```diff
@@ -473,8 +475,13 @@ TestCluster.prototype.close = function close(cb) {
 };
 
 TestCluster.prototype.createApplication =
-function createApplication(hostPort) {
+function createApplication(hostPort, bootFile) {
     var self = this;
+
+    if (bootFile === undefined) {
```
jcorbin: `|| X == null` ?

test/lib/test-cluster.js
```diff
@@ -231,7 +233,7 @@ function grow(n, callback) {
         var i = self.apps.length;
         var j = 0;
         for (; j < n; i++, j++) {
-            var app = self.createApplication('127.0.0.1:0');
+            var app = self.createApplication('127.0.0.1:0', null);
```
jcorbin: ...or was the point to have no ringpop hosts down below?

test/lib/test-cluster.js
Raynos: explicit undefined check needed here for back-compat.

test/lib/test-cluster.js
Raynos: this change silences an annoying test-only warning.

handler.js
Raynos: will fix. keeping at error for now; we will make logger.error() stop paging with umonitor in the future.

Raynos:
@kriskowal 

> can you explain why you’ve made the routing loop a remote configurable option?

During a rolling deploy half the cluster will call the new endpoint and the other half of the cluster does not implement the new endpoint.

We need to roll it out dark & then enable it once all hyperbahn workers implement `discoverAffine`

kriskowal:
@Raynos Nobody is calling this endpoint right now, and if they are, it’s getting killed. Could get rid of this complexity and just lift the ban after the deploy.

kriskowal:
:ok:

PR #211, opened 2016-01-08
Log `willSample` prepublish replacer

rf:
This requires an updated larch so it isn't ready to land yet, but it works.

Wrote a simple script to insert `willSample` checks before log lines. I messed around with using some macro systems (sweet.js and the C preprocessor) but found that neither would work. The replacer script is an inelegant solution but it's short and it works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "test-repeat": "NODE_DEBUG=autobahn node test/index.js | FORCE_COLOR=1 tap-spec; while [ $? -eq 0 ]; do NODE_DEBUG=autobahn node test/index.js | FORCE_COLOR=1 tap-spec; done;",
     "test-cover": "istanbul cover test/index.js",
     "view-cover": "opn ./coverage/index.html",
-    "prepublish": "git ls-files | grep '.js$' | grep -v 'bin/' | grep -v test | grep -v replacer.js | xargs node replacer.js"
+    "prepublish": "git ls-files | grep '.js$' | grep -v 'bin/' | grep -v test | grep -v replacer.js | xargs -L 1 node replacer.js"
   },
   "engines": {
     "node": "0.10.x"
--- a/package.json
+++ b/package.json
@@ -78,7 +78,8 @@
     "test-ci": "npm run check-licence && npm run lint -s && npm run cover",
     "test-repeat": "NODE_DEBUG=autobahn node test/index.js | FORCE_COLOR=1 tap-spec; while [ $? -eq 0 ]; do NODE_DEBUG=autobahn node test/index.js | FORCE_COLOR=1 tap-spec; done;",
     "test-cover": "istanbul cover test/index.js",
-    "view-cover": "opn ./coverage/index.html"
+    "view-cover": "opn ./coverage/index.html",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -78,7 +78,8 @@
     "test-ci": "npm run check-licence && npm run lint -s && npm run cover",
     "test-repeat": "NODE_DEBUG=autobahn node test/index.js | FORCE_COLOR=1 tap-spec; while [ $? -eq 0 ]; do NODE_DEBUG=autobahn node test/index.js | FORCE_COLOR=1 tap-spec; done;",
     "test-cover": "istanbul cover test/index.js",
-    "view-cover": "opn ./coverage/index.html"
+    "view-cover": "opn ./coverage/index.html",
+    "prepublish": "git ls-files | grep '.js$' | grep -v 'bin/' | grep -v test | grep -v replacer.js | xargs node replacer.js"
```
Raynos: We need the same prepublish in tchannel :)

Which means tchannel is coupled to larch... which means... let's just figure it out.

Raynos:
This seems reasonable.

I want to try running it and looking at a diff of the pre-process vs post-process.

Also we need to figure out what to do in terms of workflow, prepublish might have to copy into a "build/*" folder and we should change the package.json to import from there.

Then we can .gitignore build/*

Raynos:
tests fail lal.

rf:
Changed this to use publish.sh. It's real ugly but it runs the replacer, makes a temp commit, creates the package, then resets git.

publish.sh
```diff
@@ -37,7 +38,16 @@ else
     git push origin --tags
 fi
 
+# run the replacer script to make all logsites check for sampling
+git ls-files | grep '.js$' | grep -v 'bin/' | grep -v test | grep -v replacer.js | xargs node replacer.js
+
+# need a temp commit because git archive uses HEAD
+git commit -a -m 'temp commit'
```
Raynos: Run the test suite here.

rf: it'll fail bc DebugLotron everywhere (including in TChannel's test cluster or whatever) which doesn't have the right interface

Raynos:
we need to fix debug logtron.

replacer.js
```diff
@@ -0,0 +1,136 @@
...
+    msgMatches = nextLine.match(/ *(['"][^{]*['"])/);
+
+    if (!msgMatches) {
+        return null;
+    }
+
+    return msgMatches[1];
+}
+
+// Inserts an if (logger.willSample('level')) before a logsite for a particular
```
jcorbin: update comment: `willSample('level', 'msg')`

replacer.js
```diff
@@ -0,0 +1,136 @@
...
+        var matches = line.match(/( *)([_a-z\.]*logger).[a-z]*(.*)/);
+        if (!matches) {
+            throw new Error('couldn\'t match log line in replacer.js');
+        }
+        var whitespace = matches[1];
+        var logger = matches[2];
+        var restOfLine = matches[3];
+        var logMsg = findMsg(restOfLine, nextLine);
+        if (!logMsg) {
+            throw new Error('couldn\'t find log message for ' + filename + ':' + lineno);
```
jcorbin: there are a small number of semi-dynamic log sites, e.g. the recently changed https://github.com/uber/hyperbahn/blob/master/circuits.js#L202

PR #212, opened 2016-01-12
Ensure we dual-write latency stats.

Raynos:
This will give us visibility on a per-worker basis
for which hyperbahn instance has the highest latency

r: @jcorbin @rf


jcorbin:
lgtm, and yay more tests.

PR #216, opened 2016-01-21
Switch away from my local ip, test failure mode

jcorbin:
r @raynos


clients/index.js
```diff
@@ -300,6 +319,34 @@ function setupChannel(cb) {
...
+        var reason = validateHost(host);
+        if (reason !== null) {
+            nextAttempt(host, reason);
+            return;
+        }
+        self.tchannel.listen(self._port, host);
+    }
+
+    function nextAttempt(host, reason) {
+        var delay = self._getHostForTchannelAttemptBackoff * Math.pow(2, getHostAttemptCount);
```
Raynos: We should have a total time limit too. `Math.pow(2, 10)` is rather big.

We can set the total time limit to 30s or 60s or something.

test/clients/channel.js
```diff
@@ -0,0 +1,83 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+'use strict';
+
+var os = require('os');
+var path = require('path');
+var StaticConfig = require('static-config');
+var test = require('tape');
+require('leaked-handles');
```
Raynos: probably dont want to commit this.

Raynos:
I think requiring leaked-handles might create a leak...

Raynos:
Tests will probably pass if you remove leaked-handles.

clients/index.js
jcorbin: - the time limit is enforced by `_getHostForTchannelAttemptLimit`
- because of how the logic works, we never sleep for `delay * 2^10`, the last sleep is `delay *2^9`
- the total worst-case time slept (modulo fuzz) will then be `delay * 2^10 - 1` since ~17 minutes is maybe too much I'll lower the limit to 8 so that the upper bound drops to `delay * 2^8 - 1` or ~4:15

jcorbin: Correction: the upper bound with limit = `8` is `4:27.75` (taking the worst case on every random fuzz to be +5%.)

jcorbin: I've added a comment block explaining the expected runtime under defaults.

clients/index.js
```diff
@@ -276,6 +290,20 @@ function loadHostListFile(bootFile) {
     return null;
 };
 
+/* On the runtime of the localIp validation loop below:
```
jcorbin: I've added a comment block explaining the expected runtime under defaults.

clients/index.js
Raynos: We do rolling deploys and have a 2minute timeout on a worker coming up.

I'd be nice to have it bail out in <2minute.

Raynos:
lgtm.

PR #219, opened 2016-01-29
Fix circuit breaker bug from prod.

Raynos:
We were double counting the circuit breaker by calling
`circuit.onRequest()` but then noting that we could
not handleLazily and returning early.

This broke the circuit breaker, we also potentially double
count rate limiter.

In fact we do EVERYTHING again for the eager path; I think
this entire code path from "try lazy; oops eager" is going
to be brutally expensive for `ad()` and `relay-ad()`

We may have to port `ad()` and `relay-ad()` to be fully
lazy endpoints for performance reasons.

r: @rf @kriskowal @jcorbin

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -300,6 +300,16 @@ function handleLazily(conn, reqFrame) {
         return null;
     }
 
+    // use the rd (routing delegate) or the serviceName if there was no rd set
+    var serviceChannel = self.channel.subChannels[nextService];
+    if (!serviceChannel) {
+        serviceChannel = self.createServiceChannel(nextService);
+    }
+
+    if (!serviceChannel.handler.handleLazily) {
+        return false;
+    }
+
     if (self.rateLimiterEnabled) {
         var rateLimitReason = self.rateLimit(callerName, nextService);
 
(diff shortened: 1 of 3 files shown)


test/register/register-with-slow-affinity.js
```diff
@@ -57,80 +60,219 @@ allocCluster.test('register with slow affine', {
...
+function sendNRegisters(cluster, n, cb) {
+    var dummy = cluster.dummies[0];
+    var exitNodes = cluster.getExitNodes('hello-bob');
+    var counter = n;
+    var errors = [];
+
+    sendNext();
+
+    function sendNext() {
+        if (--counter === 0) {
```
kriskowal: >= 0 because paranoia

kriskowal:
:ok:

test/register/register-with-slow-affinity.js
Raynos: ಠ_ಠ

Only once...

service-proxy.js
```diff
@@ -274,7 +274,7 @@ function handleLazily(conn, reqFrame) {
         );
         // TODO: protocol error instead?
         conn.sendLazyErrorFrameForReq(reqFrame, 'BadRequest', 'failed to read serviceName');
-        return false;
```
jcorbin: oops.... dang....

jcorbin:
Good catch... this was a terrible bug.

rf:
:+1:

test/register/register-with-slow-affinity.js
kriskowal: :ok:

PR #220, opened 2016-01-30
remote-config: add changedKeys and only update if changed

Raynos:
This should make the cluster stabler and should reduce
the spikes in event loop lag when we change flipr.

r: @jcorbin @rf @kriskowal


rf:
WE WILL MAKE THE CLUSTER
STAPLER




:+1:

PR #221, opened 2016-01-30
Propagate callerName in relay-ad & disable circuits

Raynos:
This change does two things:

 - update `callerName` to be propagated.
 - Disable circuit breaker for `ad` or `relay-ad`.

After speaking with @kriskowal and @rf we realized
that the PeerReaper() will empty the entire cluster
if `relay-ad` is circuit broken.

We need to think harder about this failure mode and
what to do with it.

r: @jcorbin @rf @kriskowal


rf:
cool :+1:

PR #224, opened 2016-02-04
Generalize circuit shorting, now with visibility

jcorbin:
Add a new `"circuits.shorts"` config option which should be a `Map<String,bool>` where each string is a `"caller~service~endpoint"` tuple with simple wildcarding supported.

This is then used to drive a new `ShortedState` circuit state which is an `UnhealthyState` extension which always passes traffic.

cc @kriskowal @rf @raynos

### DIFF
--- a/circuits.js
+++ b/circuits.js
@@ -113,10 +113,10 @@ function Circuits(options) {
 }
 
 Circuits.prototype.updateShorts = function updateShorts(shorts) {
-    this.shorts = shorts;
+    this.shorts = parseShorts(shorts);
     var keys = Object.keys(this.circuitsByServiceName);
     for (var i = 0; i < keys.length; ++i) {
-        this.circuitsByServiceName[keys[i]].updateShorts(shorts);
+        this.circuitsByServiceName[keys[i]].updateShorts();
     }
 };
 
@@ -124,13 +124,18 @@ Circuits.prototype.isShorted = function isShorted(callerName, serviceName, endpo
     if (!this.shorts) {
         return false;
     }
-    return this.shorts[callerName + '~' + serviceName + '~' + endpointName]
-        || this.shorts['*~' + serviceName + '~' + endpointName]
-        || this.shorts[callerName + '~*~' + endpointName]
-        || this.shorts[callerName + '~' + serviceName + '~*']
-        || this.shorts['*~*~' + endpointName]
-        || this.shorts['*~' + serviceName + '~*']
-        || this.shorts[callerName + '~*~*'];
+
+    var byService = this.shorts[callerName] || this.shorts['*'];
+    if (!byService) {
+        return false;
+    }
+
+    var byEndpoint = byService[serviceName] || byService['*'];
+    if (!byEndpoint) {
+        return false;
+    }
+
+    return byEndpoint[endpointName] ? true : false;
 };
 
 Circuits.prototype.getCircuit = function getCircuit(callerName, serviceName, endpointName) {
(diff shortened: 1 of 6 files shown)


service-proxy.js
```diff
@@ -453,17 +453,18 @@ function handleRequest(req, buildRes) {
...
         if (!circuit.state.shouldRequest()) {
-            buildRes().sendError('Declined', 'Service is not healthy');
-            return;
+            if (!circuit.shorted) {
+                buildRes().sendError('Declined', 'Service is not healthy');
+                return;
+            }
+            self.logger.warn('short circuiting unhealthy circuit', circuit.extendLogInfo({
+            }));
+            self.statsd.increment('circuits.unhealthy.shorted', 1);
```
Raynos: useful to put `serviceName` in the stat.

service-proxy.js
```diff
@@ -453,17 +453,18 @@ function handleRequest(req, buildRes) {
...
             req.headers.cn || 'no-cn', req.serviceName, req.endpoint
         );
         if (!circuit.state.shouldRequest()) {
-            buildRes().sendError('Declined', 'Service is not healthy');
-            return;
+            if (!circuit.shorted) {
+                buildRes().sendError('Declined', 'Service is not healthy');
+                return;
+            }
+            self.logger.warn('short circuiting unhealthy circuit', circuit.extendLogInfo({
```
Raynos: please add, cn, serviceName, endpoint as well.

service-proxy.js
```diff
@@ -453,17 +453,18 @@ function handleRequest(req, buildRes) {
         serviceChannel = self.createServiceChannel(nextService);
     }
 
-    if (serviceChannel.handler.circuits &&
-        (req.serviceName !== 'hyperbahn' && (
-            req.endpoint !== 'ad' || req.endpoint !== 'relay-ad'
-        ))
-    ) {
+    if (serviceChannel.handler.circuits) {
```
Raynos: this logic should be duplicated for lazy as well.

The reason the hyperbahn gaurd only happens in eager path is because ad() is an eager endpoint handler...

Raynos:
Seems great. Especially the visibility bit.

service-proxy.js
jcorbin: aye, that's what circuit.extendLogInfo adds

service-proxy.js
jcorbin: no longer needed as I'm doing it all internally to circuits, however they're used

jcorbin:
Now with tests, ready for review.

circuits.js
```diff
@@ -97,6 +112,27 @@ function Circuits(options) {
...
+    if (!this.shorts) {
+        return false;
+    }
+    return this.shorts[callerName + '~' + serviceName + '~' + endpointName]
+        || this.shorts['*~' + serviceName + '~' + endpointName]
+        || this.shorts[callerName + '~*~' + endpointName]
+        || this.shorts[callerName + '~' + serviceName + '~*']
+        || this.shorts['*~*~' + endpointName]
+        || this.shorts['*~' + serviceName + '~*']
+        || this.shorts[callerName + '~*~*'];
```
rf: may be worth benchmarking if `this.shorts` should be `Map<string, Map<string, <Map<string, bool>>>` with a lot of conditionals, string building may be more expensive

jcorbin: `Circuits#isShorted` is called on the hot request path at most once, at circuit creation time.  Thereafter it's called only at `remoteConfig` change time (the `updateShorts` path).

jcorbin:
Updated to minimize the new overhead to circuit creation time pointed out by @rf; assumptions being that the abundantly common case will fail the first `callerName` lookup.

kriskowal:
:sheep: it

PR #225, opened 2016-02-05
Cleanup circuit observability

jcorbin:
Out of #224, claw back stats and log emitting code into the Circuit object from
ServiceProxy.

Slightly changes log messages which will require an update to analysis tooling.

May be marginally faster as it reduces the number of string concats on hot path.


Raynos:
+1

Raynos:
Think it broke tests becuase `logger.whitelist()`

PR #226, opened 2016-02-05
Less closure and event object allocation under circuits

jcorbin:
Now we have on closure for all circuits rather than than one-per-each-circuit.

r @raynos @rf


Raynos:
lgtm.

Raynos:
so good.

kriskowal:
legit

PR #227, opened 2016-02-08
Simplify circuits

jcorbin:
- drop self pattern
- make Circuit better encapsulated by injecting the root
- drop another layer of event-emission coupling
- drop unused locked states

r @raynos @rf @kriskowal


kriskowal:
:ok: This looks correct. It might be worthwhile to flatten the state objects themselves onto the circuit. We probably mis-predicted a need for extensibility here.

kriskowal:
:ok: up to this point now

PR #228, opened 2016-02-09
Further simplify circuit state change observation

jcorbin:
- no event emitter at all, just inject logger and statsd
- less checking necessary, and more string re-use

NOTE: since we no longer involve `serviceProxy.extendLogInfo` or
`channel.extendLogInfo`, the following log fields have been removed from
circuit state change logs:

```javascript
info.affineServices         = Object.keys(self.exitServices);
info.channelDestroyed       = self.destroyed;
info.channelDraining        = self.draining;
info.channelListened        = self.listened;
info.channelListening       = self.listening;
info.circuitsEnabled        = self.circuitsEnabled;
info.hostPort               = self.hostPort;
info.minPeersPerRelay       = self.minPeersPerRelay;
info.minPeersPerWorker      = self.minPeersPerWorker;
info.partialAffinityEnabled = self.partialAffinityEnabled;
info.rateLimiterEnabled     = self.rateLimiterEnabled;
```

I think this is okay and that we don't need these fields, reviewers please
double check.

r @raynos @kriskowal @rf

### DIFF
--- a/circuits.js
+++ b/circuits.js
@@ -195,7 +195,14 @@ Circuit.prototype.setState = function setState(StateType) {
         oldState.onDeactivate();
     }
 
-    this.observeTransition(state.name, oldState, state);
+    var statsPrefix = 'circuits.' + state.name;
+    this.root.statsd.increment(statsPrefix + '.total', 1);
+    this.root.statsd.increment(statsPrefix + this.byCallerStatSuffix, 1);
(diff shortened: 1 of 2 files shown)


kriskowal:
The lost log fields are okay.

kriskowal:
:ok:

PR #229, opened 2016-02-11
Fix remote config change encapsulation

jcorbin:
So that each topic funtion does its own noop checking, bounding the complexit
of the change dispatcher, and allowing for more nuanced handling of changed
keys in a few places.

r @rf @kriskowal

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -462,10 +462,7 @@ function onRemoteConfigUpdate(changedKeys, forceUpdate) {
     self.updateReapPeersPeriod(hasChanged, forceUpdate);
     self.updatePrunePeersPeriod(hasChanged, forceUpdate);
     self.updatePartialAffinityEnabled(hasChanged, forceUpdate);
-
-    if (forceUpdate || hasChanged['relay.maximum-ttl']) {
-        self.setMaximumRelayTTL();
-    }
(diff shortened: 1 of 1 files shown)


rf:
:+1:

clients/index.js
```diff
@@ -564,98 +514,130 @@ ApplicationClients.prototype.updateLazyHandling = function updateLazyHandling()
...
+
+    if (forceUpdate || hasChanged['log.reservoir.size']) {
         var size = self.remoteConfig.get('log.reservoir.size', 100);
-        var interval = self.remoteConfig.get('log.reservoir.flushInterval', 50);
+        self.logReservoir.setSize(size);
+    }
 
+    if (forceUpdate || hasChanged['log.reservoir.flushInterval']) {
+        var interval = self.remoteConfig.get('log.reservoir.flushInterval', 50);
         self.logReservoir.setFlushInterval(interval);
```
kriskowal: @rf please let us know if setFlushInterval depends on a following setSize to effect the change.

kriskowal:
Before and after appear to be consistent. :ok: as long as the reservoir configuration properties are actually independent.

clients/index.js
rf: Nope, this should be fine

PR #231, opened 2016-02-12
Ensure correctness in unadvertise test

jcorbin:
Previously this test didn't handle well the case where there weren't any
connections to be closed, due to tacitly assuming that connections get opened
immediately.

Furthermore it wasn't straightforward to fix this as we were previously
depending on the non-immediate callback of untilAllInConnsRemoved to actually
wait for the unadvertise RPC to finish, rather than actually fencing on its
callback.

r @kriskowal @rf


kriskowal:
:ok:

rf:
:+1:

PR #235, opened 2016-02-15
Fix unadvertise test untilExitsConnected utility

jcorbin:
- only done once we have in-connections from all...
- ...which requires blocking for identified event (since otherwise the
  connection is as-yet unpeered, and not seen by `forEachConn`)
- drop prior `appIndex` logic that is redundant with the cluster exits check
  (since all exits are in `cluster.apps`, just not the converse)

r @rayson @rf @kriskowal

### DIFF
--- a/test/hyperbahn-client/unadvertise.js
+++ b/test/hyperbahn-client/unadvertise.js
@@ -250,9 +250,13 @@ function untilExitsConnected(cluster, remote, callback) {
         });
         var gotExits = Object.keys(got).length;
         if (gotExits >= numExists) {
-            remote.channel.connectionEvent.removeListener(checkConns);
-            callback();
+            finish();
         }
+    }
+
+    function finish() {
+        remote.channel.connectionEvent.removeListener(checkConns);
+        callback();
     }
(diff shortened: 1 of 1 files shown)


test/hyperbahn-client/unadvertise.js
```diff
@@ -235,25 +235,34 @@ function runTests(HyperbahnCluster) {
...
-            var appIndex = cluster.hostPortList.indexOf(peer.hostPort);
-            if (appIndex >= 0) {
-                if (exits[peer.hostPort] !== undefined) {
-                    got[peer.hostPort] = true;
-                }
+            if (exits[peer.hostPort] !== undefined && conn.direction === 'in') {
+                got[peer.hostPort] = true;
             }
         });
         var gotExits = Object.keys(got).length;
```
kriskowal: It does not matter in this case, but you can avoid Object.keys calls for each event by populating an object initially, deleting keys, and then using a fast isObjectEmpty check instead.

kriskowal:
:ok:

PR #237, opened 2016-02-16
TestCluster borg untilExits{Connected,Disconnected}

jcorbin:
- move the hyperbahn-client unad utilities into `TestCluster` methods
- use `untilExitsConnected` in `createRemote` to explicate the current status quo
  before deferring `peer.connectTo` time

r @raynos @rf


Raynos:
lgtm.

PR #240, opened 2016-02-17
Change circuit to return Unhealthy

Raynos:
This changes the CircuitBreaker to return an Unhealth
error frame instead of a Declined error frame.

This will:

 - Stop retries on circuit broken; to avoid retry storm
 - Make graphs a bit nicer because Unhealthy is only
	used by circuit breaker.

r: @rf @kriskowal @jcorbin

### DIFF
--- a/circuits.js
+++ b/circuits.js
@@ -37,10 +37,17 @@ function EndpointCircuits(root) {
     this.circuitsByEndpointName = {};
 }
 
-EndpointCircuits.prototype.updateShorts = function updateShorts(shorts) {
+EndpointCircuits.prototype.updateShorts = function updateShorts() {
     var keys = Object.keys(this.circuitsByEndpointName);
     for (var i = 0; i < keys.length; ++i) {
         this.circuitsByEndpointName[keys[i]].updateShorted();
+    }
+};
+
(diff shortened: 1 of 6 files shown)


jcorbin:
Please make it configurable in a way similar to how circuit shorting is configured: on `cn~sn~arg1` tuples, with simple wildcarding.  In same push, it will serve to shift the boundary away from `func (Circuit) shouldRequest() bool` to `func (Circuit) getRequestError() (code, message string)`.

rf:
lgtm

circuits.js
```diff
@@ -98,7 +97,8 @@ function Circuits(options) {
     this.statsd = options.statsd;
     this.circuitsByServiceName = {};
     this.config = options.config || {};
-    this.shorts = null;
+    this.shorts = options.shorts;
```
jcorbin: || null

jcorbin:
Recording our offline convo: please make it actually flippable per circuit, not just for all circuits.

jcorbin:
This looks :ok: to me now; ideally we'd split out:
- simplify the shorts config (originally I had it setup to thread thru, but then backed out to referencing off root only)
- change the circuit api to return error frame / reject with it
- the rest?

But I can live with this as is if you're pressed for time.

PR #246, opened 2016-02-18
WIP: call update{ServiceNodes,ExitNodes} after changeTo{Exit,Forward}

jcorbin:
Unsure if this, or a less heavy handed change is necessary to get parital
affinity working.


Raynos:
I think this is needed, but we should keep an eye on it. Can we add a custom logline here so we can see how often this happens.

This should be fine except for the CPU runaway case :(

jcorbin:
Aye, stats on changes and updates would be great, will add.

Currently the only affect of `updateServiceNodes` is under the partial affinity feature, so noop until then.

PR #248, opened 2016-02-19
service-proxy: emit recvd stat if rejecting frame

Raynos:
We had a bug where customers saw more errors then
recvd.

This is because the recvd stat is emitted by either:

 - RelayHandler
 - EndpointHandler

When we reject a request due to rate limiter or
circuit breaker we increment latency & error but do
not increment recvd

r: @jcorbin @rf @kriskowal


rf:
cool. :+1: once travis

PR #249, opened 2016-02-22
Cleanup ServiceProxy#refreshServicePeerPartially for perf analysis

jcorbin:
In most recent flame graph under canary test:
- this method took ~19% of process time
- most of that 19% was due to an anonymous function that was the alternate to freshenPartialPeer

Hopefully this will provide us more clarity once / if we get down to tuning this path.

r @rf @Raynos


kriskowal:
:ok:

Raynos:
lgtm.

PR #251, opened 2016-02-23
Correctly handle rpsLimits set to zero...

jcorbin:
...by effectively disabling rate limits.

I noticed that our test suite at least was provoking rate limiting for total
rps limit set to 0.  Whether or not the tests should be running with rps limit
0, I feel this is more correct implementation.

r @raynos @rf


Raynos:
lgtm.

Raynos:
lgtm.

PR #257, opened 2016-02-24
Add deregister (unad) support to TestCluster remotes

jcorbin:


kriskowal:
:ok:

jcorbin:
Well... adding this was the straw that broke the "function complexity lint rule's back"... so I'm pursuing https://github.com/uber/hyperbahn/pull/258 first.

PR #250, opened 2016-02-23
Add peer churn test

jcorbin:
Sketch:
- single client, 10 peers
- can send, hit the 10 peers
- new 10 peers
- can send, hit the 20 peers
- shutdown the first 10 peers (half of them gracefully by unad, the others just destroy)
- create blackhole fixtures in their place
- assert that the blackholes receive no incoming connections over several periods
- can send, hit the second 10 peers only
- assert that the blackholes received no incoming connections while sending traffic

PR #259, opened 2016-02-24
Do incremental work under ServiceProxy#removeServicePeer

jcorbin:
Use partialRange.removeWorker rather than sorting all service workers.


jcorbin:
Going to block and rebase on https://github.com/uber/hyperbahn/pull/250 since it adds more coverage to the affected path.

Raynos:
I trust the method does what it says on the tin.

jcorbin:
it does, and it's already used by the other way that we remove workers under the peer reaper: https://github.com/uber/hyperbahn/blob/master/service-proxy.js#L1471

jcorbin:
Amusingly enough, this method did not work as advertised; fixed in #267.

PR #261, opened 2016-02-25
Add staleness checking for partial affinity workers

jcorbin:
Rather than always create peers by side-effect, check the partialRange.workers
against actual serviceChannel peeers, remove stale ones, and log loudly about
it.


jcorbin:
Dear reviewers,

I'm going to split out isomorphic work first, let me know if you want the "don't create" semantic change separated from the "staleness check" change.

Raynos:
Looks fine.

Raynos:
It looks ok.

PR #267, opened 2016-02-26
Fix bug in PartialRange#removeWorker

jcorbin:
Previously it never did any such thing.

Bug revealed by the upcoming staleness auditing (#261) under the recently added
peer churn test with partial affinity enabled.


Raynos:
nice find. lgtm.

PR #269, opened 2016-02-26
WIP: pivot connectedServicePeers into partialRange.connectedWorkers

jcorbin:


jcorbin:
may be a small perf win for partial affinity

PR #271, opened 2016-02-29
WIP: test peer churn under sustained forwarding load

jcorbin:


jcorbin:
it's also amusing to observe that the endpoint delay that I added will probably allow to test at higher scale of other parameters, even in travis, since the cpu is less contended due to the slow endpoint causing the test machine to smear over time more.

PR #273, opened 2016-03-01
make test-cluster multi process parallel safe

Raynos:
RTAPI runs multiple processes with multiple concurrent
hyperbahn clusters.

r: @rf @kriskowal @jcorbin

### DIFF
--- a/test/lib/remote-config-file.js
+++ b/test/lib/remote-config-file.js
@@ -33,7 +33,10 @@ function RemoteConfigFile(name) {
     }
 
     var self = this;
-    name = name || 'UNKNOWN[' + process.pid + ']';
+
+    name = name || 'UNKNOWN';
+    name += '[' + process.pid + ']';
+
     self.filePath = path.join('/tmp',
                               'remote_config_' + name + '.json');
 }


kriskowal:
:ok:

test/lib/remote-config-file.js
```diff
@@ -33,9 +33,15 @@ function RemoteConfigFile(name) {
...
 
     var self = this;
-    name = name || 'UNKNOWN[' + process.pid + ']';
-    self.filePath = path.join('/tmp',
-                              'remote_config_' + name + '.json');
+
+    if (!name) {
+        name = 'UNKNOWN[' + process.pid + ']';
+    } else {
+        name += '[' + process.pid + ']';
```
jcorbin: how about the simpler:
```
var name = name || 'UNKNOWN';
name += '[' + process.pid + ']';
```

test/lib/remote-config-file.js
```diff
@@ -33,9 +33,15 @@ function RemoteConfigFile(name) {
...
+    }
+    self.filePath = path.join(
```
jcorbin: no change on these lines, y u churn

jcorbin:
sure... but could be simpler

Raynos:
Thanks @jcorbin I made it less churny :)

PR #275, opened 2016-03-01
Factor out AffinityChange#removeWorker

jcorbin:
Catches a missing `delete this.connectedPeers[worker];` that was causing stale
logs (but not provoking turnip connections) for the even churn peers that do a
graceful unadvertise shutdown.


Raynos:
lgtm.

PR #277, opened 2016-03-07
skip another test to avoid flaps

Raynos:
This avoids a flap in travis. This test got more flappy
recently :S 

r: @jcorbin @rf @kriskowal


kriskowal:
:ok:

PR #278, opened 2016-03-10
allow recvd requests for debugging

Raynos:
This will allow us to update the `hyperbahn-worker` dashboard
and see all recvd requests for that istance.

This will help debug latency issues and event loop lag issues

r: @jcorbin @rf

### DIFF
--- a/clients/batch-stats.js
+++ b/clients/batch-stats.js
@@ -60,12 +60,12 @@ function handleStat(stat) {
     }
 
     if (
-        stat.name === 'tchannel.inbound.calls.latency'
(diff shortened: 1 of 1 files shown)


kriskowal:
:ok:

rf:
fien :+1:

jcorbin:
:ok: -ish but it's debt that this needs to be a code change; should really have a list/lookup table of global stats and per worker stats.... if you try to PR this code path again, I may consider blocking until you make it configurable (aka this one's free, but now you know ;-)

PR #280, opened 2016-03-11
clients: add projectScope for statsd & ringpop

Raynos:
This allows us to configure the string that we use for
our stats and ringpop based on external factors.

We want to keep the kafka string seperately configurable
from the stats / ringpop string.

r: @kris @jcorbin @rf


kriskowal:
It might be more clear to break the "project" config into more meaningful names, e.g., "kafkaName", "statsName", "ringpopName".

rf:
or `projectPrefix` but I don't really care just stop paging me plz :+1:

Raynos:
NO U TRAVIS.

PR #281, opened 2016-03-14
service-proxy: loop over service keys instead

Raynos:
This is a nasty bug where we did not delete peers
in the reaper.

We accidentally "loop" over an object...

This also fixes an elk issue where we are exploding
their key space with {serviceName} dimension.

r: @rf @kriskowal @jcorbin

### DIFF
--- a/service-proxy.js
+++ b/service-proxy.js
@@ -1400,8 +1400,9 @@ function reapSinglePeer(hostPort, serviceNames, now) {
         peer.clearDrain('superceded by peer reap');
     }
 
-    for (var i = 0; i < serviceNames.length; i++) {
-        var serviceName = serviceNames[i];
+    var serviceKeys = Object.keys(serviceNames);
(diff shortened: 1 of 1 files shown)


service-proxy.js
```diff
@@ -1400,8 +1400,9 @@ function reapSinglePeer(hostPort, serviceNames, now) {
         peer.clearDrain('superceded by peer reap');
     }
 
-    for (var i = 0; i < serviceNames.length; i++) {
-        var serviceName = serviceNames[i];
+    var serviceKeys = Object.keys(serviceNames);
```
jcorbin: more minimal to rename the arg and then have `var serviceNames = Object.keys(someArg)`... the values are what, last refresh / advertise time? is it useful to log those too?

jcorbin:
:ok: but please consider comment

service-proxy.js
Raynos: not sure if values are useful to log; never really looked at the data.

PR #282, opened 2016-03-16
Add noindex to avoid log indexing

Raynos:
We have a few fields where the keyPath has a huge
cardinality. By postfixing `_noindex` to a key name in 
our logs we will not blow up the elk search indexing for
our logs.

r: @jcorbin @rf @kriskowal


rf:
:+1:

kriskowal:
:ok:

jcorbin:
:ok: would prefer maybe `counters_noindex: {service: {...}, edge: {...}}`

PR #285, opened 2016-04-19
also emit stats to global client

blampe:
Hello hi what is this.

@prashantv @Raynos @sloppyfocus

### DIFF
--- a/test/clients/statsd.js
+++ b/test/clients/statsd.js
@@ -61,7 +61,8 @@ test('statsd can talk to a server', function t(assert) {
                 'my-app..' + os.hostname().split('.')[0] +
                     '.some-stat:1|c\n',
                 'my-app.per-worker..' + os.hostname().split('.')[0] +
-                    '.test-hyperbahn.some-stat:1|c\n'
+                    '.test-hyperbahn.some-stat:1|c\n',
+                'my-app.all-workers.some-stat:1|c\n'
             ]);
 
             server.close();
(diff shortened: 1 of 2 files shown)


Raynos:
lgtm.

We should do this for `timing()` and all other methods as well.

prashantv:
+1, should also add this all the other methods

test/clients/statsd.js
```diff
@@ -61,7 +61,8 @@ test('statsd can talk to a server', function t(assert) {
                 'my-app..' + os.hostname().split('.')[0] +
                     '.some-stat:1|c\n',
                 'my-app.per-worker..' + os.hostname().split('.')[0] +
-                    '.test-hyperbahn.some-stat:1|c\n'
+                    '.test-hyperbahn.some-stat:1|c\n',
+                'my-app.all-workers.some-stat:1|c\n'
```
Raynos: I suspect that because you call globalClient.increment() first that you have to make this line the first in the array rather then last.

Raynos: Or not... guess it doesnt matter :D tests pass.

Raynos:
lgtm. merge at will.

blampe:
Very flakey tests :(

@Raynos after I land what's the release/deploy process?

Raynos:
@blampe 

Just run `bash publish.sh`; since it requires `npm publish` permissions I can run it on your behalf or you can login in as the `uber` npm user.

blampe:
@Raynos I'll probably take you up on that, let's plan on releasing/deploying Monday. I'd like to look over your shoulder while you do it.

Raynos:
Running

```
bash ./publish.sh patch
```

Created `hyperbahn@2.14.36`

PR #288, opened 2016-05-16
Upgrade to TChannel 3.7.0

kriskowal:
Upgrades TChannel for production validation. Works around an uncaught exception when sending error frames on a reclaimed lazy relay request. Allows optional arguments flag can be threaded through TChannel, but Hyperbahn should make no use of this at this time.


kriskowal:
r @Raynos @jcorbin

kriskowal:
Received offline 👍  from @jcorbin

PR #289, opened 2016-06-22
remove per-worker timings

blampe:
Per-worker latency timings represent a majority of our metrics but rolling these up to the host level is sufficient.

@kriskowal @sloppyfocus


kriskowal:
@Raynos please sign. CI will pass soon.

Raynos:
This is wrong location for this fix. Look in batch-stats instead.

https://github.com/uber/hyperbahn/blob/master/clients/batch-stats.js

kriskowal:
@Raynos i.e., axe this line? https://github.com/uber/hyperbahn/blob/master/clients/batch-stats.js#L67

blampe:
@kriskowal I'll let you take it from here :)

clients/dual-statsd.js
```diff
@@ -114,7 +114,6 @@ DualStatsd.prototype.timing = function timing(name, time) {
 
     self.globalClient.timing(name, time);
     self.perServerClient.timing(name, time);
-    self.perWorkerClient.timing(name, time);
```
Raynos: We want to keep this line.

This lets us track event loop lag per worker.

clients/batch-stats.js
```diff
@@ -64,7 +64,6 @@ function handleStat(stat) {
         stat.name === 'tchannel.inbound.calls.recvd'
     ) {
         self.writeToStatsd(self.statsd.globalClient, stat);
-        self.writeToStatsd(self.statsd.perWorkerClient, stat);
```
Raynos: We want to keep this line.

I recommend we keep `recvd` per worker as it allows us to understand what QPS is going through a single HOT worker. this is a common debugging technique based on our alerts.

I recommend we remove `latency` for most service with the exception of the `hyperbahn` service being whitelisted, tracking the hyperbahn latency on a per-worker basis allows us to debug any issues with advertise.

Raynos:
lgtm.

PR #298, opened 2016-08-04
Support individual shard repositioning

kriskowal:
- [x] Add iHops to EgressNodes computations
- [x] Thread iHops configuration through remote config
- [ ] Test

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -671,6 +671,11 @@ ApplicationClients.prototype.updateKValues = function updateKValues(hasChanged, 
         }
     }
 
+    if (forceUpdate || hasChanged['iHops.services']) {
+        self.egressNodes.iHopsForServiceName = self.remoteConfig.get('iHops.services', {});
+        changed = true;
+    }
+
     if (changed) {
         self.serviceProxy.updateServiceChannels();
     }
--- a/egress-nodes.js
+++ b/egress-nodes.js
@@ -40,6 +40,7 @@ function EgressNodes(options) {
     self.defaultKValue = options.defaultKValue;
 
     self.kValueForServiceName = {};
+    self.iHopsForServiceName = {}; // iHopsForServiceName[serviceName][i] -> i'
 
     // Surface the membership changed event (for use in particular by service
     // proxies).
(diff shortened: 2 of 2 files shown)


kriskowal:
r @jcorbin @rf

clients/index.js
```diff
@@ -671,6 +671,11 @@ ApplicationClients.prototype.updateKValues = function updateKValues(hasChanged,
         }
     }
 
+    if (forceUpdate || hasChanged['iHops.services']) {
```
jcorbin: sUch cApitalization iS oNly aSking fOr pRoblems

egress-nodes.js
```diff
@@ -40,6 +40,7 @@ function EgressNodes(options) {
     self.defaultKValue = options.defaultKValue;
 
     self.kValueForServiceName = {};
+    self.iHopsForServiceName = {}; // iHopsForServiceName[serviceName][i] -> i'
```
jcorbin: `i'` for "eye prime" is a bit opaque as far as comments go... but sure

egress-nodes.js
```diff
@@ -88,9 +97,13 @@ EgressNodes.prototype.exitsFor = function exitsFor(serviceName) {
     );
 
     var k = self.kValueFor(serviceName);
+    var iHops = self.iHopsFor(serviceName);
     // Object<hostPort: String, Array<lookupKey: String>>
     var exitNodes = Object.create(null);
     for (var i = 0; i < k; i++) {
+        if (iHops[i]) {
```
jcorbin: ` !== undefined`, or do you mean to disallow mapping to `0`?

egress-nodes.js
```diff
@@ -117,8 +130,12 @@ EgressNodes.prototype.isExitFor = function isExitFor(serviceName) {
     );
 
     var k = self.kValueFor(serviceName);
+    var iHops = self.iHopsFor(serviceName);
     var me = self.ringpop.whoami();
     for (var i = 0; i < k; i++) {
+        if (iHops[i]) {
```
jcorbin: same wrt `0`

egress-nodes.js
rf: we can't map to 0 (or anything < k) but it's not purposeful

rf:
lgtm

jcorbin:
I'm 🆗 with this as long as @kriskowal  agrees wrt zaro.

kriskowal:
I do agree wrt zaro hops. moar hops oar bettar. you’d never hop to a value less than k in practice, since it would just double up on a host that's already affine. However, there’s no harm in erring on the side of caution, so I’ve made the change.

jcorbin:
Your battling for zaro hath inspired an LG™

PR #303, opened 2016-09-02
No longer under active development

blampe:
@kriskowal


DavidTPate:
I'd love to hear what alternative you are using instead. I'm guessing REST calls all over the place instead.

Raynos:
lgtm.

@DavidTPate We use many things, this includes tchannel & http. We use either a router ( hyperbahn, haproxy ) or we use direct connect ( using host files and client-side load balancing ). Honestly, a lot of these things used in production are not under active development. 

We have teams working on new solutions that will be used in the future like https://github.com/yarpc. We are slowly ramping down our internal hyperbahn traffic and migrating it to a new internal tchannel router.

PR #306, opened 2016-09-28
Ringpop upgrade

kriskowal:
r @Raynos


kriskowal:
recommend minor?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hyperbahn",
-  "version": "2.17.3",
+  "version": "2.17.3-ringpop-upgrade",
```
jcorbin: why not just patch

Raynos:
Lgtm, minor seems reasonable.

PR #307, opened 2016-09-30
Add subchannel update damper

kriskowal:
This adds a configurable option for ServiceProxy to debounce requests to update service channels by a minimum period (default 0). The period can be altered by an external agent (remote configuration).

The updateServiceChannels function is not covered by our tests currently. None of our tests appear to exercise ring changes.

Without configuring the min update period, this retains the current behavior, except using setTimeout instead of setImmediately, effectively introducing an implicit minimum delay of 4ms.

r @Raynos


Raynos:
lgtm.

Recommend hooking up `minUpdatePeriod` to flipr before deploying.

Recommend spamming restart in travis until you get the "lucky green".

kriskowal:
@Raynos Added remote config.

Raynos:
Still lgtm :P

PR #310, opened 2016-10-03
Configure process reporter handle interval

### LINKED ISSUE #308: Configure ProcessReporter `handleInterval`
Currently 6% of CPU is reporting handles

Change from `1000` -> `30000`

kriskowal:
Fixes #308 

This change makes the process reporter run every 30 seconds instead of 1ce per second in production.

r @Raynos  @blampe

### DIFF
--- a/clients/index.js
+++ b/clients/index.js
@@ -143,7 +143,7 @@ function ApplicationClients(options) {
 
     self.processReporter = ProcessReporter({
         statsd: self.statsd,
-        handleInterval: config.get('clients.process-reporter.handle-interval', 1000)
+        handleInterval: config.get('clients.process-reporter.handle-interval')
     });
 
     // This is dead code; really really soon.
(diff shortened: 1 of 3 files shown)


clients/index.js
```diff
@@ -142,7 +142,8 @@ function ApplicationClients(options) {
     });
 
     self.processReporter = ProcessReporter({
-        statsd: self.statsd
+        statsd: self.statsd,
+        handleInterval: config.get('clients.process-reporter.handle-interval', 1000)
```
Raynos: `config.get()` takes one argument. The second argument is ignored.

The semantics of `config.get()` are that if the key does not exist throw an exception.

Just remove the `1000`.

kriskowal: I’ll add a development default as well then.

Raynos:
lgtm.

PR #314, opened 2016-10-24
Ease the connection stale period to 10s

kriskowal:
r @raynos @jcorbin


Raynos:
This is supposed to be 1500ms of no activity on the Connection and at least two timeouts.

Since most timeouts range from 1, 5, or 10s. Changing this to a 10 second period of no activity and at least two timeouts seems reasonable.

jcorbin:
😋

PR #315, opened 2017-05-23
Make test logger optional

MadanThangavelu:
This helps reduce noise during testing

### DIFF
--- a/test/lib/test-cluster.js
+++ b/test/lib/test-cluster.js
@@ -133,7 +133,7 @@ function TestCluster(opts) {
     self.timers = null; // Set whenever a channel is created
 
     self.tchannelJSON = TChannelJSON();
-    self.logger = DebugLogtron('autobahn');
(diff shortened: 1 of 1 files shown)


Matt-Esch (approved):

Matt-Esch:
I would make sure that the test cluster doesn't require being able to read the debug logger buffer.

test/lib/test-cluster.js
```diff
@@ -133,7 +133,7 @@ function TestCluster(opts) {
     self.timers = null; // Set whenever a channel is created
 
     self.tchannelJSON = TChannelJSON();
-    self.logger = DebugLogtron('autobahn');
+    self.logger = opts.logger || DebugLogtron('autobahn');
```
Raynos, line 136: you can also do

```
DebugLogtron('autobahn', {
    enabled: opts.loggerEnabled
});
```

Setting `enabled` to false, silences everything except `error` and `fatal`.

